Issue #2, opened 2018-02-23
Cannot register tunnel


mholt:
While doing some refactoring in a branch, I finally got it to a running state but now I have this error. I checked out the original branch (master) just to be sure, and the error still appears (I added a single print line to show the error message):

```
$ go build && ./cloudflare-warp --hostname my.hostname --hello-world --no-autoupdate
INFO[0000] Proxying tunnel requests to https://127.0.0.1:53287 
INFO[0000] Starting Hello World server at 127.0.0.1:53287 
INFO[0000] Starting metrics server                       addr="127.0.0.1:53288"
ERRO[0000] handle return: received return for unknown question id=0  subsystem=rpc-transport
INFO[0000] Connected to LAX                             
pogs: insert @0xc793e50592935b4a: can't insert into pogs.RegistrationOptions: pogs.RegistrationOptions has unknown field PoolName, maps to poolName
ERRO[0000] Cannot register                              
INFO[0000] Retrying in 1s seconds
```

Does this have something to do with an inconsistency between the client and the backend maybe? I actually have no idea.

mholt:
Commenting out the fields `PoolName` and `IsAutoupdated` in `pogs.RegistrationOptions` struct resolves the problem. I just don't comprehend at what cost. :) It's working now, with that change. Will leave open for advisement.

chungthuang:
Hi mholt, thanks for reporting the issue! We will tunnelrpc.capnp.go.

chungthuang:
tunnelrpc.capnp.go has been updated. You shouldn't need to comment out `PoolName` and `IsAutoupdated` in `pogs.RegistrationOptions` anymore.

PR #3, opened 2018-02-24
Initial WIP refactoring logic out of main package

### LINKED ISSUE #1: Refactor to separate command from library functions
I would like to propose refactoring the repository so that most of the logic is moved out of the `main` package and into a new package called `warp` or `cfwarp`.

That way, the `main` package can call library functions to log in, start the server/tunnel, etc, as would a Caddy plugin or k8s ingress controller, without duplicating code.

I'm starting on this locally already, and if I can get it working I'll submit a PR.

mholt:
(WIP -- not ready to merge. But I will try to keep the pushes here functioning so you can use it.)

This change moves some logic out of the main package and into a new `warp` package so that other clients can use this as a library.

Specifically, the login code that gets credentials and the code that starts a server has been moved over (mostly). Most of the additions are related to a rather sizable struct type that shuttles configuration settings into the `StartServer()` function.

Some TODOs:

- [ ] Improve logging; it'd be nice if the library didn't _require_ logrus (for example, Caddy doesn't use logrus and has logs of its own it could pass in) -- can we make an interface and pass in some "general" logger instead?

- [ ] Test login more

- [ ] Test startups/shutdowns more. I'm not sure if this related to my changes, but after a few trial runs I couldn't get the tunnel to register anymore -- no errors are produced, but accessing my warp hostname in a browser shows "503 Service Unavailable The origin has been unregistered from Cloudflare Warp" -- even after switching back to the master branch and compiling with no changes, I couldn't get that error to go away, despite no errors on the client-side.

- [ ] Fix the issue reported in #2. Right now this PR just comments out the offending lines as a workaround.

Closes #1

### DIFF
--- a/warp/server.go
+++ b/warp/server.go
@@ -1,0 +1,186 @@
+package warp
+
+import (
+	"crypto/tls"
+	"encoding/hex"
+	"fmt"
+	"io/ioutil"
+	"math/rand"
+	"net"
+	"net/http"
+	"os"
+	"time"
+
+	"github.com/cloudflare/cloudflare-warp/origin"
+	"github.com/cloudflare/cloudflare-warp/tlsconfig"
+	tunnelpogs "github.com/cloudflare/cloudflare-warp/tunnelrpc/pogs"
+	"github.com/cloudflare/cloudflare-warp/validation"
+	homedir "github.com/mitchellh/go-homedir"
+	"github.com/sirupsen/logrus"
+)
+
+// StartServer starts a warp proxy server with the given configuration.
+// It blocks indefinitely.
+func StartServer(cfg ServerConfig) error {
+	hostname, err := validation.ValidateHostname(cfg.Hostname)
+	if err != nil {
+		return err
+	}
+	if cfg.ClientID == "" {
+		cfg.ClientID = generateRandomClientID()
+	}
+
+	cfg.Tags = append(cfg.Tags, tunnelpogs.Tag{Name: "ID", Value: cfg.ClientID})
+
+	cfg.ServerURL, err = validation.ValidateUrl(cfg.ServerURL)
+	if err != nil {
+		return fmt.Errorf("validating server URL: %v", err)
+	}
+
+	// Check that the user has acquired a certificate using the log in command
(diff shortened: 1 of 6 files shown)


dknecht:
There should never be a 503.  If so thats a bug.

mholt:
For anyone following along, the Caddy plugin is now up at: https://github.com/caddyserver/cloudflare-warp-plugin

elithrar:
@mholt - thanks for contributing!

* RE: a Logger interface - yes, definitely prefer this. Directly logging in libraries isn't a good experience, so accepting a minimal interface - e.g. 

```
type Logger interface {
	Log(keyvals ...interface{}) error
}
```

* RE: 503's - as Dane said, this shouldn't happen. We've put some mechanisms in place to prevent this, but you are correct in that it's a propagation delay (Tunnel backend -> provisions other services elsewhere -> needs those to services to push their config to the edge quickly).

mholt:
@dknecht Resolved merge conflicts. Finally got my computer back from the shop so I'm mostly back in business (after I get over this jet lag)!

Can we possibly get this reviewed/merged sooner rather than later? Resolving merge conflicts when the warp client is updated is a bit tedious and risky for me to do alone.

As for the logging TODO, I don't really think it has to be resolved in this PR; seems like a separate thing.

warp/server.go
```diff
@@ -0,0 +1,182 @@
...
+		ProtocolLogger:    cfg.ProtoLogger,
+		Logger:            cfg.Logger,
+		IsAutoupdated:     cfg.IsAutoupdated,
+	}
+
+	// blocking
+	return origin.StartTunnelDaemon(tunnelConfig, cfg.ShutdownChan, cfg.ConnectedChan)
+}
+
+func fileExists(path string) (bool, error) {
```
sssilver: Is this function redundant with `if _, err := os.Stat(path); err == nil { ... }`?

mholt: For the purposes of whether a file _exists_, yes, more or less -- although you'd want to check if `!os.IsNotExist(err)` instead of a blanket `err == nil`, since there might be other reasons that statting a file fails.

PR #11, opened 2018-04-10
Makes origin.TunnelMetrics public again

ntfrnzn:
Test code must avoid the prometheus static initialization, and
can do this by creating a dummy TunnelMetrics object, provided
it is public.


ntfrnzn:
Resolved at 77e7c240c0d838b3a377954a412b528d9070714f

PR #24, opened 2018-04-28
HTTP timeout set to wrong value.

yon2004:
Because the defaultTimeout is defined as a constant you do not need to multiply it by time.Second again otherwise you end up with some stupid timeout as seen here. https://play.golang.org/p/Dl7rU_n_8Ej


vavrusa:
Thanks! I'll add it to the next release.

chungthuang:
Thank you for the contribution! The fix was released in version 2018.5.0.

PR #28, opened 2018-05-16
Fix POST requests

### LINKED ISSUE #5: 411 Length Required
Client proxy get no problem , when I try to post some data , show me 

Length Required

A request of the requested method POST requires a valid Content-length.
Apache Server at  Port 80

Config.yml   As below

hostname: www.**.org
url: http://192.168.88.188

seroy:
Fixes #5


mthird:
looks like the github is having trouble with the merge due to the repo change.

@sssilver what would it take to get this (or something like this) merged?

PR #35, opened 2018-07-04
Metrics labels

ntfrnzn:
Request for comments, will squash/rebase before continuing.

Adds a set of metrics label keys to the TunnelHandler and uses the TunnelHandler to send all metrics updated.  The base metrics labels are defined by the application and passed through and made available when the TunnelHandler is created.

See also branch at https://github.com/cloudflare/cloudflare-ingress-controller/tree/metrics.

### DIFF
--- a/cmd/cloudflared/main.go
+++ b/cmd/cloudflared/main.go
@@ -482,10 +482,11 @@ func startServer(c *cli.Context, shutdownC, graceShutdownC chan struct{}) error 
 		return err
 	}
 
+	metricsLabels := map[string]string{"application": "cloudflared"}
 	wg.Add(1)
 	go func() {
 		defer wg.Done()
-		errC <- origin.StartTunnelDaemon(tunnelConfig, graceShutdownC, connectedSignal)
+		errC <- origin.StartTunnelDaemon(tunnelConfig, graceShutdownC, connectedSignal, metricsLabels)
 	}()
 
 	return waitToShutdown(&wg, errC, shutdownC, graceShutdownC, c.Duration("grace-period"))
--- a/origin/metrics.go
+++ b/origin/metrics.go
@@ -1,6 +1,8 @@
 package origin
 
 import (
+	"hash/fnv"
+	"strconv"
 	"sync"
 	"time"
 
@@ -28,25 +30,25 @@ type muxerMetrics struct {
 }
 
 type TunnelMetrics struct {
-	haConnections     prometheus.Gauge
-	totalRequests     prometheus.Counter
-	requestsPerTunnel *prometheus.CounterVec
+	haConnections prometheus.Gauge
+	totalRequests prometheus.Counter
+	requests      *prometheus.CounterVec
 	// concurrentRequestsLock is a mutex for concurrentRequests and maxConcurrentRequests
 	concurrentRequestsLock      sync.Mutex
 	concurrentRequestsPerTunnel *prometheus.GaugeVec
-	// concurrentRequests records count of concurrent requests for each tunnel
-	concurrentRequests             map[string]uint64
+	// concurrentRequests records count of concurrent requests for each tunnel, keyed by hash of label values
+	concurrentRequests             map[uint64]uint64
 	maxConcurrentRequestsPerTunnel *prometheus.GaugeVec
-	// concurrentRequests records max count of concurrent requests for each tunnel
-	maxConcurrentRequests map[string]uint64
+	// concurrentRequests records max count of concurrent requests for each tunnel, keyed by hash of label values
+	maxConcurrentRequests map[uint64]uint64
 	timerRetries          prometheus.Gauge
-	responseByCode        *prometheus.CounterVec
-	responseCodePerTunnel *prometheus.CounterVec
-	serverLocations       *prometheus.GaugeVec
+
+	reponses        *prometheus.CounterVec
+	serverLocations *prometheus.GaugeVec
 	// locationLock is a mutex for oldServerLocations
 	locationLock sync.Mutex
 	// oldServerLocations stores the last server the tunnel was connected to
-	oldServerLocations map[string]string
+	oldServerLocations map[uint64]string
 
 	muxerMetrics *muxerMetrics
 }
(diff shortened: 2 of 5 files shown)


origin/tunnel.go
```diff
@@ -542,6 +542,7 @@ func (h *TunnelHandler) ServeStream(stream *h2mux.MuxedStream) error {
 	h.logRequest(req, cfRay, lbProbe)
 	if websocket.IsWebSocketUpgrade(req) {
 		conn, response, err := websocket.ClientConnect(req, h.tlsConfig)
+		h.logger.WithFields(log.Fields{"connectionID": h.connectionID, "status": response.StatusCode}).Info("incrementResponses")
```
chungthuang: Response is logged in line 555 and line 568.

cmd/cloudflared/main.go
```diff
@@ -482,10 +482,11 @@ func startServer(c *cli.Context, shutdownC, graceShutdownC chan struct{}) error
 		return err
 	}
 
+	metricsLabels := map[string]string{"application": "cloudflared"}
```
chungthuang: What's the reason to add this label when the metrics names are prefix with argotunnel?

origin/metrics.go
```diff
@@ -28,162 +30,162 @@ type muxerMetrics struct {
...
 
 	muxerMetrics *muxerMetrics
 }
 
-func newMuxerMetrics() *muxerMetrics {
+func newMuxerMetrics(baseMetricsLabels []string) *muxerMetrics {
 	rtt := prometheus.NewGaugeVec(
 		prometheus.GaugeOpts{
-			Name: "rtt",
+			Name: "argotunnel_rtt",
```
chungthuang: We should make argotunnel the namespace of all the metrics in this file.

origin/metrics.go
```diff
@@ -206,120 +208,120 @@ func newMuxerMetrics() *muxerMetrics {
...
 
-	responseByCode := prometheus.NewCounterVec(
-		prometheus.CounterOpts{
-			Name: "response_by_code",
-			Help: "Count of responses by HTTP status code",
-		},
-		[]string{"status_code"},
-	)
-	prometheus.MustRegister(responseByCode)
+	// responseByCode := prometheus.NewCounterVec(
```
chungthuang: This can be deleted.

origin/metrics.go
```diff
@@ -206,120 +208,120 @@ func newMuxerMetrics() *muxerMetrics {
...
-		muxerMetrics:                   newMuxerMetrics(),
+
+		responses:          responses,
+		serverLocations:    serverLocations,
+		oldServerLocations: make(map[uint64]string),
+		muxerMetrics:       newMuxerMetrics(baseMetricLabels),
+	}
+}
+
+func hashLabelValues(labelValues []string) uint64 {
```
chungthuang: ConnectionID is unique per tunnel. Is there other reason to use the hash value of labels as key?

cmd/cloudflared/main.go
ntfrnzn: It's there mostly as an example of how to add additional labels, but it could distinguish cloudflared from other applications (the ingress controller) that are using the library.

origin/metrics.go
ntfrnzn: seems to be '0' everywhere, apparently StartTunnelDaemon sets it https://github.com/cloudflare/cloudflared/blob/master/origin/tunnel.go#L135  Bug?

origin/tunnel.go
```diff
@@ -516,6 +542,8 @@ func (h *TunnelHandler) ServeStream(stream *h2mux.MuxedStream) error {
...
```
ntfrnzn: this pattern, `h.xxx(h.yyy(h.z))` seems like a code smell and suggests that more of the logic be moved from metrics into the TunnelHandler

origin/tunnel.go
ntfrnzn: +1

ntfrnzn: Although I see that errors are logged differently, without the response.  And also assumes that all errors are status code 502.  I think it makes more sense to log the response in all cases.  I agree that the logging is redundant

origin/metrics.go
chungthuang: ConnectionID is 0 in that case because config.HAConnections is set to 1.

cmd/cloudflared/main.go
chungthuang: Is namespace not enough to distinguish cloudflared from other applications?

origin/metrics.go
sssilver: https://godoc.org/github.com/prometheus/client_golang/prometheus#Opts has `Namespace` -- we should use this instead (@chungthuang is right)

origin/metrics.go
```diff
@@ -28,162 +30,162 @@ type muxerMetrics struct {
...
-func newMuxerMetrics() *muxerMetrics {
+func newMuxerMetrics(baseMetricsLabels []string) *muxerMetrics {
 	rtt := prometheus.NewGaugeVec(
 		prometheus.GaugeOpts{
-			Name: "rtt",
+			Name: "argotunnel_rtt",
 			Help: "Round-trip time in millisecond",
 		},
-		[]string{"connection_id"},
+		append(baseMetricsLabels, "connection_id"),
```
rishabh-bector: Do we have to `append()` for each metric? Can we just do it once and pass it to all of them?

origin/metrics.go
```diff
@@ -28,162 +30,162 @@ type muxerMetrics struct {
...
 	)
 	prometheus.MustRegister(rtt)
 
 	rttMin := prometheus.NewGaugeVec(
 		prometheus.GaugeOpts{
-			Name: "rtt_min",
+			Name: "argotunnel_rtt_min",
 			Help: "Shortest round-trip time in millisecond",
 		},
```
sssilver: Instead of the above, I would:
```
rttMin := NewGaugeVec(
    "rtt_min", 
    "Shortest round-trip time in milliseconds", 
    append(baseMetricsLabels, "connection_id")
)

// ...

func NewGaugeVec(name, help string, metricsLabels []string) prometheus.GaugeVec {
    const ARGO_TUNNEL_NAMESPACE = "argo_tunnel"

    return prometheus.NewGaugeVec(
        prometheus.GaugeOpts{
            Name: name,
            Namespace: ARGO_TUNNEL_NAMESPACE,
            Help: help,
        },
        metricsLabels
    )
}
```

This helps keep the code DRY, and leaves less room for people to make mistakes.

PR #65, opened 2018-12-28
make http transport aware of proxy from envvar

kols:


kols:
Is this repo still alive?

sssilver:
Yes, very much so -- thanks for the contribution and sorry for taking this long!

PR #76, opened 2019-03-07
Add rdp as a supported protocol in URL validation

joelmiles:


validation/validation.go
```diff
@@ -14,7 +14,7 @@ import (
 
 const defaultScheme = "http"
 
-var supportedProtocol = [2]string{"http", "https"}
+var supportedProtocol = [2]string{"http", "https", "rdp"}
```
sssilver: `[2]` should probably just be `[]`, and it would be nice if `supportedProtocol` was called `supportedProtocols` (plural) at this point

joelmiles: Fixed both.

PR #78, opened 2019-03-19
adjust macOS launchctl to auto-restart/keepalive

mrjfox:
I would like cloudflared to auto restart as i find the process often crashes with no error logs reported. Here is a PR for this simple change.

### DIFF
--- a/cmd/cloudflared/macos_service.go
+++ b/cmd/cloudflared/macos_service.go
@@ -55,10 +55,7 @@ func newLaunchdTemplate(installPath, stdoutPath, stderrPath string) *ServiceTemp
 		<key>StandardErrorPath</key>
 		<string>%s</string>
 		<key>KeepAlive</key>
-		<dict>
-			<key>SuccessfulExit</key>
-			<false/>
-		</dict>
+		<true/>
 		<key>ThrottleInterval</key>
 		<integer>20</integer>
 	</dict>


mrjfox:
version in use: cloudflared version 2019.3.1 (built 2019-03-18-2312 UTC)

OzzieOrca:
Seems to crash on me when changing networks. Haven't been able to find the logs.

nickvollmar:
This request makes sense. Thank you for writing in and making a PR with this fix.

I believe the root cause of this issue is https://github.com/cloudflare/cloudflared/issues/96

We consider some kinds of errors to be "permanent" –– they can't be fixed just by restarting cloudflared. For example: when the user's Cloudflare account doesn't have the required billing set up. We don't want the service manager to retry `cloudflared` on "permanent" errors, as retrying would not accomplish anything. But other errors are "temporary," and if `cloudflared` encounters a "temporary" error, we do want it to be restarted by the service manager.

The idiomatic way to distinguish between "temporary" and "permanent" errors seems to be: 
1. have the process `exit(0)` on a "permanent" error, nonzero on "temporary"
2. configure the service manager not to retry on exit code 0

Clearly we're not doing (1) properly, because we're exiting successfully on a "temporary" error.

Potential fixes are:
* stop doing (2) as well. This stops the false positives, by treating all errors as "temporary." But if the error is truly "permanent," `cloudflared` may thrash and bog down the system (unless macOS launchctl has an exponential backoff?). That is the fix implemented in this PR.
* do (1) properly. This would also stop the false positives, and avoids the risk of the process thrashing.

In conclusion, I'd prefer if we did (1) properly in https://github.com/cloudflare/cloudflared/issues/96 –– does that make sense?

cblecker:
This change isn't a good idea. If the application quits for a known successful reason (graceful shutdown, system rebooting, software update), launchd will still try to restart the application immediately.

PR #84, opened 2019-04-11
configure git to access to http(s) based repos behind CloudFlare Access

mustafakirimli:
example command: `cloudflared access gitConfig https://git.site.com`

### DIFF
--- a/cmd/cloudflared/access/cmd.go
+++ b/cmd/cloudflared/access/cmd.go
@@ -73,6 +73,14 @@ func Commands() []*cli.Command {
 					SkipFlagParsing: true,
 				},
 				{
+					Name:   "gitConfig",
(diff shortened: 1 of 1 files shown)


mustafakirimli:
It requires min git version 2.9.1

mustafakirimli:
Following shortcut can be used to have wrapper around git commands in case if token expires it can trigger cloudflared access call. 
```
#!/bin/bash
git () {
  if [ "$1" == "clone" ]; then
    cloudflared access gitConfig $2
  elif [ -d "${PWD}/.git" ]; then
    repoUrl=$(command git config --get remote.origin.url)
    cloudflared access gitConfig $repoUrl
  fi
  command git $@
}
```

PR #89, opened 2019-04-16
Add dockerfile to make an official image

sevki:
Closes #62

PR #90, opened 2019-04-29
Build a docker container

jbergstroem:
The build phase makes the assumption that any capnproto-files are pre-processed before building in-docker. I personally prefer "add-on-change" for protos over add-on-compile since that may lead to a dirty build.

Chose to add this as a PR even though there's already an open one. Feel free to close if we should focus on one PR with a slightly different direction instead.

Finally, my comment in the previous PR (#89) set out to clone inside the container which likely is a safer approach for release builds, but copying tree allows us to do "developer"/in-transit builds.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,12 @@
+FROM golang:1.12 as builder
+WORKDIR /go/src/github.com/cloudflare/cloudflared/
+RUN apt-get update && apt-get install -y --no-install-recommends upx
+# Run after `apt-get update` to improve rebuild scenarios
+COPY . .
+RUN make cloudflared
+RUN upx --no-progress cloudflared
+
+FROM gcr.io/distroless/base
+COPY --from=builder /go/src/github.com/cloudflare/cloudflared/cloudflared /usr/local/bin/
+ENTRYPOINT ["cloudflared", "--no-autoupdate"]
+CMD ["version"]
--- a/Makefile
+++ b/Makefile
@@ -31,6 +31,10 @@ clean:
 cloudflared: tunnel-deps
 	go build -v $(VERSION_FLAGS) $(IMPORT_PATH)/cmd/cloudflared
 
+.PHONY: container
+container:
+	docker build -t cloudflare/cloudflared:"$(VERSION)" .
+
 .PHONY: test
 test: vet
 	go test -v -race $(VERSION_FLAGS) ./...


Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM golang:1.12-alpine as builder
+
+WORKDIR /go/src/github.com/cloudflare/cloudflared/
+COPY . .
+
+RUN apk add --no-cache gcc musl-dev git upx ca-certificates
```
nickvollmar: Should we run `apk update && apk add`? I usually see that done in Dockerfiles, to pick up changes to the registry between the base image build and the Dockerfile build.

Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM golang:1.12-alpine as builder
+
+WORKDIR /go/src/github.com/cloudflare/cloudflared/
+COPY . .
```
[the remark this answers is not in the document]
jbergstroem: Good point!

Dockerfile
jbergstroem: [Upstream](https://github.com/docker-library/golang/blob/d6715f0733f5282836cad2867a0a2b107fc35147/1.12/alpine3.9/Dockerfile) doesn't seem to do it and in general I'm not a fan of doing so because it means we're sliding the updates between builds. This should mean we layer our container properly.

Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+
+RUN apk add --no-cache gcc musl-dev git upx ca-certificates
+RUN VERSION=$(git describe --tags --always --dirty="-dev") && \
+    DATE=$(date -u '+%Y-%m-%dT%H:%MZ') && \
+    go build -ldflags "-X main.Version=\"${VERSION}\" -X main.BuildTime=\"${DATE}\"" \
```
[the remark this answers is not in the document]
jbergstroem: Thanks for the suggestion, will incorporate.

Dockerfile
jbergstroem: Since this is somewhat subjective I'll leave it for you to decide what you prefer.

cblecker:
@jbergstroem If you use the debian-based golang image as the builder, you can just use the `make cloudflared` rule and not include all the extra golang flags in the Dockerfile. This makes the docker build more consistent with what would be built on a local system, and doesn't make a difference to the end image as you discard the builder anyways.

Additionally, if you use [`gcr.io/distroless/base`](https://github.com/GoogleContainerTools/distroless) for the final image then the image size is smaller and has the smallest available attack surface.

Example:
```dockerfile
FROM golang:1.12 as builder
WORKDIR /go/src/github.com/cloudflare/cloudflared/
COPY . .
RUN apt-get update && apt-get install -y --no-install-recommends upx
RUN make cloudflared && upx --no-progress cloudflared

###

FROM gcr.io/distroless/base
COPY --from=builder /go/src/github.com/cloudflare/cloudflared/cloudflared /
ENTRYPOINT ["/cloudflared"]
```

Results in:
```
$ docker images cloudflared
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
cloudflared         latest              5597453d2a6e        5 minutes ago       32.1MB
```

jbergstroem:
> @jbergstroem If you use the debian-based golang image as the builder

I can change; up to upstream wrt what they prefer really.

nickvollmar:
Very interesting! I had not heard of distroless until now. Thank you for the suggestion, @cblecker. I agree that reducing the attack surface area is valuable.

Surprisingly, the Distroless image size seems to be a bit _bigger_ than the Alpine image.
```
$ docker images --filter=reference='cloudflared:*'
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
cloudflared         distroless          1728712c0b9e        2 minutes ago       32.1MB
cloudflared         alpine              6689ad34f5a6        7 minutes ago       21.3MB
```

My first reflex was that I'd mixed up the tags, but I double- and triple-checked that this wasn't the case...

My next assumption was that the Alpine executable was considerably smaller. But no, it's actually a bit larger than the one built by Debian for Distroless (15.4M golang:1.12-alpine, 15.2M golang:1.12).

It seems like there's something to be investigated/tweaked here. If Distroless is really... distroless, then my instinct tells me it should be smaller than any distro-based image...

For the sake of publishing a Docker image, my reflex is to merge this PR as-is and investigate Distroless later.

cblecker:
@nickvollmar the difference is in the libc libraries. Alpine has Busybox + APK + muslc, while distroless is the full glibc library.

More details: https://medium.com/@dwdraju/distroless-is-for-security-if-not-for-size-6eac789f695f

The other distinct advantage is because it’s the full glibc library, you can use the debian based image for the build step with make, so you keep the build commands identical and in a single place

cblecker:
The other thing I noticed that could make the cloudflared image smaller is stripping the debugging information from the release binary (adding `-s -w` to the ldflags). But that's probably best for a separate PR

jbergstroem:
Switched to debian as base and rebased on top of latest master.

> The other thing I noticed that could make the cloudflared image smaller is stripping the debugging information from the release binary (adding `-s -w` to the ldflags). But that's probably best for a separate PR

Yes; happy to open another PR that would add a `-debug` docker tag and slim the release as a result if people think it would be used. I don't mind the few extra megs since containers are compressed and downloaded as such over the wire.

Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM golang:1.12 as builder
+
+WORKDIR /go/src/github.com/cloudflare/cloudflared/
+COPY . .
+RUN apt-get update && apt-get install -y --no-install-recommends upx
+RUN make cloudflared
+RUN upx --no-progress cloudflared
+
+FROM grc.io/distroless/base
```
cblecker: I think this is a typo..
```suggestion
FROM gcr.io/distroless/base
```

jbergstroem: Thanks for pointing that out. I managed to add some noise while modifying the commit message with `git commit --amend`. Fixed.

jbergstroem:
A friend of mine brought up that `base` from distroless also adds openssl which would be unecessary. With that aid I think we can stay with this for convenience reasons (since the alternative would be going back to static linking again).

Dockerfile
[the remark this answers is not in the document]
nickvollmar: Would y'all mind swapping the `RUN apt-get` and the `COPY` instructions, to improve ergonomics when rebuilding the image?

Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+
+WORKDIR /go/src/github.com/cloudflare/cloudflared/
+COPY . .
+RUN apt-get update && apt-get install -y --no-install-recommends upx
+RUN make cloudflared
+RUN upx --no-progress cloudflared
+
+FROM gcr.io/distroless/base
+COPY --from=builder /go/src/github.com/cloudflare/cloudflared/cloudflared /usr/local/bin/
+ENTRYPOINT ["cloudflared"]
```
nickvollmar: cloudflared has an auto-update feature; it can be disabled by running with the `--no-autoupdate` flag. Do you think that's worth adding to the entrypoint? I think that might make sense: as a user of a container, I expect it not to rebuild itself at runtime...

nickvollmar:
Oh, oops, I'm bad at Github. I meant to include the comment https://github.com/cloudflare/cloudflared/pull/90/files#r290885688 in that last message.

Dockerfile
cblecker: Yeah, I think this would make sense, especially considering we're versioning containers based off of versions/tags. We would want it to stay whatever version we explicitly create.

jbergstroem: > @nickvollmar said: cloudflared has an auto-update feature; it can be disabled by running with the --no-autoupdate flag.

I was leaning on this output from the cloudflared log: 
```console
INFO[0000] cloudflared will not automatically update when run from the shell. To enable auto-updates, run cloudflared as a service: https://developers.cloudflare.com/argo-tunnel/reference/service/ 
```

..but there's definitely a point in being explicit here, especially if behavior changes in `cloudflared`.

Dockerfile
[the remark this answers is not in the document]
jbergstroem: Done

nickvollmar:
Thanks!

@sssilver Could you merge this when you get a chance?

jbergstroem:
Since there's not enough comments in this PR I took the opportunity of summarizing the size differences between distroless flavors (`-strip` incorporates changes from #102):
```console
$  docker images --filter=reference='cloudflared:*'
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
cloudflared         static-strip        a0ef335ac719        17 seconds ago      8.92MB
cloudflared         static              793ad1c667a0        2 minutes ago       14.6MB
cloudflared         base                561100fb7ba6        4 minutes ago       32.1MB
```

PR #95, opened 2019-05-06
Use Go Modules for dependency tracking

paulquerna-okta:
- Uses Go Modules for dependency tracking.
- Recommend reviewing individual commits to understand.
- Includes https://github.com/cloudflare/cloudflared/pull/94 for sanity, since it lets you use the latest CoreDNS release.


go.mod
```diff
@@ -0,0 +1,42 @@
...
+	github.com/facebookgo/subset v0.0.0-20150612182917-8dac2c3c4870 // indirect
+	github.com/getsentry/raven-go v0.2.0
+	github.com/golang-collections/collections v0.0.0-20130729185459-604e922904d3
+	github.com/google/uuid v1.1.1
+	github.com/gorilla/mux v1.7.1
+	github.com/gorilla/websocket v1.4.0
+	github.com/kylelemons/godebug v1.1.0 // indirect
+	github.com/lib/pq v1.1.1
+	github.com/mattn/go-colorable v0.1.1
+	github.com/miekg/dns v1.1.9
```
nickvollmar: Due to https://github.com/miekg/dns/issues/941 we pinned this dependency at commit version `6da3249dfb57fbaa16efafcd8744cee8809d80cd`:
https://github.com/cloudflare/cloudflared/commit/781e1fef7bafec91b242682a94b9f37719a0e788

If I understand go modules correctly, this can be accomplished by running `go get github.com/miekg/dns@6da3249dfb57fbaa16efafcd8744cee8809d80cd` and committing the result –– would you mind doing that?

paulquerna-okta: @nickvollmar https://github.com/cloudflare/cloudflared/commit/bab7583a97e8f7776640889f8b823f63a77e709a

Bumped it off of `6da3249dfb57fbaa16efafcd8744cee8809d80cd` to "=1.1.8", was that a mistake? (this is the current `master`)

In this specific case, `go get` with a pinned version would cause go modules to downgrade `coredns` dependency, This is because coredns v1.5.0 requires a newer version of `dns`.  So, to override a pin when a dependency declares a version, you need to use `replace`, like this:

```
go mod edit -replace github.com/miekg/dns=github.com/miekg/dns@6da3249dfb57fbaa16efafcd8744cee8809d80cd
```

Done in 9a3a9fc4a09d6e5abbf3f6615f51935106253bdf and 5785bb46207da525578188bfbbcd28c1a5d240dc

PR #100, opened 2019-05-25
Fix macOS launch daemon

shinvou:
On macOS you'd start the proxy with '[sudo] cloudflared proxy-dns'. This launch daemon didn't start the proxy since the argument 'proxy-dns' was missing. This commit adds the argument to the launch daemon and now starting the proxy on system startup works as expected.


cblecker:
@shinvou Shouldn't this (`proxy-dns: true`) be in your cloudflared config file instead? It's possible to run cloudflared in argo tunnel mode and not run DNS proxying, so this shouldn't necessarily be forced in service creation.

shinvou:
You are right, I wasn't aware of the configuration file. Thanks, closing.

PR #102, opened 2019-06-02
Add -w -s to ldflags when compiling for release

cblecker:
By default, when compiling cloudflared via make, we can strip the symbol
table and DWARF debugging information from the binary. This results in
an approximately 10Mb reduction in file size, with no impact on
execution.

If you need to run cloudflared through a debugger and need the
DWARF/symbol table information, you can run `make cloudflared DEBUG=true` to
use the previous behaviour.


jbergstroem:
If we land this post container PR, perhaps we can add the `-debug` logic as part of it? Also, I think you should add `-debug` to VERSION.

cblecker:
@nickvollmar @adamchalmers This has been rebased. Is this still okay to merge?

ytbigvinny27 (approved):

PR #104, opened 2019-06-03
Access/ssh-config: generate consistent ssh_config

herver:
* Prefer text/template over html/template to avoid unwanted HTML encoding
* Use templated "cfpipe-{{.Hostname}}"
* Fix IdentityFile suffix in "cfpipe" to use key generated by "ssh-gen"


acmacalister:
@herver Thanks for this! Great minds think alike as we fixed this internally with an almost identical PR. We are working on getting a release with this fix out right now. Thanks for again for the contribution. We appreciate it.

PR #125, opened 2019-08-12
rename rdp to tcp

mustafakirimli:


cmd/cloudflared/tunnel/cmd.go
```diff
@@ -546,7 +546,7 @@ func hostnameFromURI(uri string) string {
 	switch u.Scheme {
 	case "ssh":
 		return addPortIfMissing(u, 22)
-	case "rdp":
+	case "tcp":
 		return addPortIfMissing(u, 3389)
```
sodabrew: This port, 3389, is for the RDP protocol: https://en.wikipedia.org/wiki/Remote_Desktop_Protocol

sssilver: Thanks for the PR, @mustafakirimli! Before we review the code (thanks @sodabrew!), do you mind creating an issue in the Github project outlining what this change is accomplishing and why?

sodabrew:
Aha, I see what you're getting at now. It doesn't even make sense that `rdp` is implemented as an alias of `ssh` but with a different default port. As you add in this PR, `tcp` must be accompanied by a port number. It's not an alias of `ssh` because none of these other subcommands made any sense with a general `tcp` service. Suggest to split `ssh` and `tcp` apart.

For backwards compatibility sake, `rdp` should have to be equal to `tcp` + `port 3389`

mustafakirimli:
> Thanks for the PR, @mustafakirimli! Before we review the code (thanks @sodabrew!), do you mind creating an issue in the Github project outlining what this change is accomplishing and why?

Kindly find the issue below, 
@sssilver 
https://github.com/cloudflare/cloudflared/issues/130

judge2020:
Keeping RDP as its own protocol might help in the future if/when Argo Tunnel supports proxying UDP, at which time an `rdp://` protocol would mean having a tunnel set up for both UDP and TCP port 3389.

daltoniam:
Thanks for the PR! This is now officially supported: https://developers.cloudflare.com/access/other-protocols/tcp-guide/

PR #126, opened 2019-08-12
refactor(docker): optimize Dockerfile

hazcod:
Remove obsolete upx binary compression
Run as unprivileged user
Use debian10 rootless base image for latest glibc libraries


jbergstroem:
> Remove obsolete upx binary compression

Could you elaborate on the rationale here? No significant savings? Startup speed?

hazcod:
@jbergstroem : I see no rationale in compressing cloudflared in a container.
The binary is already pretty small, and can be made even smaller by removing debug symbols. (https://github.com/cloudflare/cloudflared/pulls)

It only adds another level of abstraction (unpack in memory -> execute) and some small startup delay.

jbergstroem:
> It only adds another level of abstraction (unpack in memory -> execute) and some small startup delay.
Out of curiosity - is this measurable the same way the [size savings are measurable](https://github.com/cloudflare/cloudflared/pull/90#issuecomment-499249971)?

hazcod:
ping @nickvollmar

hazcod:
Can we merge this @nickvollmar @adamchalmers ?

PR #139, opened 2019-10-07
Do not allow TLS versions less than 1.2 in DoH client

NFM-8:
Added explicit MinVersion for TLS configuration.


Electroid:
What's the reasoning for this change?

moderation:
Pre TLS 1.2 not secure. Is being disabled in all browsers.

Issue #98, opened 2019-05-13
2019.4.0: Memory leak


jbergstroem:
Noticed a memory leak on a gitlab self-hosted instance over the weekend. Systemd respawned cloudflared so "all good" (sans the downtime). 

- This is debian buster (Linux gitlab-ng 4.9.0-8-amd64 SMP Debian 4.9.144-3.1 (2019-02-19) x86_64 GNU/Linux
- Was running 2019.4.0 (have updated since)

Unfortunately, Azure doesn't graph or collect memory usage for host so I can't say how fast this grew. Unfortunately number two, since cloudflared seems to zap logs on restart (edit: seems to be default journalctl behavior on debian) I can't really say what was going on at that point.

Full dmesg output:
```
[Fri May 10 20:02:04 2019] gitlab-workhors invoked oom-killer: gfp_mask=0x24201ca(GFP_HIGHUSER_MOVABLE|__GFP_COLD), nodemask=0, order=0, oom_score_adj=0
[Fri May 10 20:02:04 2019] gitlab-workhors cpuset=/ mems_allowed=0
[Fri May 10 20:02:04 2019] CPU: 2 PID: 39380 Comm: gitlab-workhors Not tainted 4.9.0-8-amd64 #1 Debian 4.9.144-3.1
[Fri May 10 20:02:04 2019] Hardware name: Microsoft Corporation Virtual Machine/Virtual Machine, BIOS 090007  06/02/2017
[Fri May 10 20:02:04 2019]  0000000000000000 ffffffffbbb34524 ffffabb14bfd7c28 ffff8bf52e5fc840
[Fri May 10 20:02:04 2019]  ffffffffbba07610 0000000000000000 0000000000000000 0000000c00000001
[Fri May 10 20:02:04 2019]  ffff8bf40f556340 ffffffffbb98b577 0000004200000000 6ac9738c427756f1
[Fri May 10 20:02:04 2019] Call Trace:
[Fri May 10 20:02:04 2019]  [<ffffffffbbb34524>] ? dump_stack+0x5c/0x78
[Fri May 10 20:02:04 2019]  [<ffffffffbba07610>] ? dump_header+0x78/0x1fd
[Fri May 10 20:02:04 2019]  [<ffffffffbb98b577>] ? get_page_from_freelist+0x3f7/0xb20
[Fri May 10 20:02:04 2019]  [<ffffffffbb9874ea>] ? oom_kill_process+0x21a/0x3e0
[...]

PR #142, opened 2019-10-12
Fix excessive memory usage streaming large files

### LINKED ISSUE #129: 2019.8.1: Tunnel memory leak
Using cloudflared 2019.8.1 with a free argo tunnel and downloading a 1gb test file from a static file server used about 1.4 GB RSS for cloudflared which was then killed by the OOM killer, since this was on a 2GB DigitalOcean droplet.

Test Setup:

* Static file server from https://gist.github.com/paulmach/7271283
* 1 GB test file from http://speedtest-fra1.digitalocean.com/1gb.test

```sh
# on server
go build serve.go
# put 1gb.test in folder webroot
./serve -d webroot &
cloudflared tunnel --url http://localhost:8100
# on client
wget -O /dev/null https://example-tunnel.trycloudflare.com/1gb.test
```

System Info:
OS: Ubuntu 18.04.3 x86_64 Linux 5.0.0-25
CPU: Xeon E5-2650L v3 @ 1.80GHz (1 core)
RAM: 2GB
Cloudflated Version: 2019.8.1
GOOS: linux, GOVersion: go1.12.7, GoArch: amd64

Installed from official .deb package.

Problem might be related to #98.

felixbuenemann:
This drops the default size auf the h2mux write buffer from 512 MB to 1 MB.
This massively reduces memory usage, since each stream has its own buffer.

Fixes #129

### DIFF
--- a/h2mux/h2mux.go
+++ b/h2mux/h2mux.go
@@ -19,7 +19,7 @@ const (
 	maxWindowSize            uint32        = (1 << 31) - 1 // 2^31-1 = 2147483647, max window size in http2 spec
 	defaultTimeout           time.Duration = 5 * time.Second
 	defaultRetries           uint64        = 5
-	defaultWriteBufferMaxLen int           = 1024 * 1024 * 512 // 500mb
+	defaultWriteBufferMaxLen int           = 1024 * 1024 // 1mb
 
 	SettingMuxerMagic http2.SettingID = 0x42db
 	MuxerMagicOrigin  uint32          = 0xa2e43c8b


felixbuenemann:
Btw. I chose 1 MB because that's twice the buffer size used by CopyBuffer to copy the reponse body.

felixbuenemann:
Most likely this also fixes #98. The oom-killer logs in the issue show 1.5 GB RSS for cloudflared.

sssilver:
Thanks for contributing!

PR #149, opened 2019-11-12
Fix Docker build failure

khawaga:
Currently Docker containers fail to build on version 2019.11.0, this merge request makes the following changes to the Dockerfile:

- Sets `GO111MODULE=on` which ensures that Go module support is activated. This resolves the following errors during Docker builds:

> build flag -mod=vendor only valid when using modules
make: *** [Makefile:32: cloudflared] Error 1

- Disables cgo by setting `CGO_ENABLED=0` which gives us a static binary, this resolves dependency issues which resulted in following error during runtime:

> cloudflared: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by cloudflared)

- Ensures target OS `GOOS` is set to Linux for Docker builds.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,7 @@
 FROM golang:1.12 as builder
 ENV GO111MODULE=on
 ENV CGO_ENABLED=0
+ENV GOOS=linux
 WORKDIR /go/src/github.com/cloudflare/cloudflared/
 RUN apt-get update && apt-get install -y --no-install-recommends upx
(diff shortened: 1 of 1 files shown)


mw866:
Can someone review & merge the PR? 
I'm seeing the same error in my build on Dockerhub.
https://hub.docker.com/repository/registry-1.docker.io/mw866/cloudflared/builds/bfaa375e-3a67-46dc-990b-0825de2c3446

PR #151, opened 2019-11-14
support for multiarch docker manifest

klutchell:
merged from https://github.com/klutchell/cloudflared
requires 'docker login' prior to 'make manifest' so the image can be pushed


klutchell:
By using 'scratch' instead of 'distroless' we can easily support multiple architectures, not just x86_64.

In this commit the Dockerfile appears to be normal since the cross-compiling is handled in the Makefile with [docker buildx](https://github.com/docker/buildx).

https://github.com/klutchell/cloudflared-1/blob/da8be8e873097255cfebaa1e207d9f84d5100c34/Makefile#L40-L49

DWSR:
I'd love to see an official ARM image built so that I can run this on my Raspberry Pi on Docker.

klutchell:
@DWSR feel free to try out https://github.com/klutchell/cloudflared in the meantime and provide any feedback there. It supports a number of platforms including arm32v7 and arm64.

mw866:
@klutchell I can see this has been fixed in  #184.

mw866:
> I'd love to see an official ARM image built so that I can run this on my Raspberry Pi on Docker.

@DWSR If you prefer not to wait for the official ARM64 docker image, you can build yours based on the official source. Here is how I did mine: https://hub.docker.com/r/mw866/cloudflared
It works on my Raspberry Pi 4B without any issues.

Issue #141, opened 2019-10-08
panic: runtime error for long lived request


ssalinas:
👋 we are testing argo tunnel for eventual use with production level traffic. While playing around with a load testing service we found we could repeatedly crash cloudflared with a specific request.

Context:

- argo tunnel is running as a sidecar process to nginx in kubernetes (not the ingress controller, just a container running cloudflared). Nginx then proxies to a java dropwizard service as its upstream
- In cloudflare this is set up as:
  - pool: ngx-argo-loadtest
  - load balancer: ngx-argo-loadtest.(domain1).net
  - load test code is hitting loadtest.(domain2).com which is a CNAME to ngx-argo-loadtest.(domain1).net . Healthchecks pass for this setup and simple requests to other endpoints can be made with no issue
- The load test hits an endpoint on loadtest.(domain2).com which proxies to the dropwizard service that then returns a random bunch of bytes of varying size

Result:

A curl to this endpoint actually succeeds:
```
*   Trying 104.17.247.168...
* TCP_NODELAY set
* Connected to loadtest.hubapiqa.com (104.17.247.168) port 443 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* Cipher selection: ALL:!EXPORT:!EXPORT40:!EXPORT56:!aNULL:!LOW:!RC4:@STRENGTH
* successfully set certificate verify locations:
*   CAfile: /etc/ssl/cert.pem
  CApath: none
* TLSv1.2 (OUT), TLS handshake, Client hello (1):
* TLSv1.2 (IN), TLS handshake, Server hello (2):
* TLSv1.2 (IN), TLS handshake, Certificate (11):
* TLSv1.2 (IN), TLS handshake, Server key exchange (12):
* TLSv1.2 (IN), TLS handshake, Server finished (14):
* TLSv1.2 (OUT), TLS handshake, Client key exchange (16):
* TLSv1.2 (OUT), TLS change cipher, Client hello (1):
* TLSv1.2 (OUT), TLS handshake, Finished (20):
* TLSv1.2 (IN), TLS change cipher, Client hello (1):
* TLSv1.2 (IN), TLS handshake, Finished (20):
* SSL connection using TLSv1.2 / ECDHE-ECDSA-CHACHA20-POLY1305
* ALPN, server accepted to use h2
* Server certificate:
*  subject: OU=Domain Control Validated; OU=PositiveSSL Multi-Domain; CN=ssl826607.cloudflaressl.com
[...]

PR #154, opened 2019-11-26
bug(cloudflared): nil pointer deference on h2DictWriter Close()

pims:
Unlike other h2DictWriter methods, the Close() method does check whether
w.comp is nil.

This PR adds a check for non nil compressor before attempting to close

Bug: #141


sssilver:
Thanks for the contribution!

PR #157, opened 2019-12-10
Warn users about websockets being a requirement for SSH access

AlexanderEkdahl:
If WebSockets are disabled to the origin the WebSocket headers will be silently stripped and any attempt to access resources over SSH will fail with `websocket: bad handshake`. This PR improves that error message.

This is not mentioned anywhere in any documentation that I could find. 

This PR also removes closing the response body as it is not necessary as per https://github.com/cloudflare/cloudflared/blob/master/websocket/websocket.go#L72-L73

### DIFF
--- a/carrier/carrier.go
+++ b/carrier/carrier.go
@@ -37,13 +37,6 @@ func (c *StdinoutStream) Write(p []byte) (int, error) {
 	return os.Stdout.Write(p)
 }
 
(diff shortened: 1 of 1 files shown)


AlexanderEkdahl:
Thinking more about it I'm not sure if this is a good idea as there currently is no way of differentiating this from the tunnel simply being down. The docs should still mention WebSockets being a requirement though!

PR #158, opened 2019-12-17
chore(dependabot): add config.yml

hazcod:
This adds dependabot integration which is integrated in GitHub.
Doing so will automatically create pull requests (including changelog/commits) whenever a go dependency is updated or the golang Docker image is updated.


sudarshan-reddy:
While this is a good idea, I think we should close this because this code base sees more sync in a bitbucket mirror.

Issue #133, opened 2019-09-10
Systemd update timer broken


Tugzrida:
After noticing that my install hadn't updated for a while, I went looking and found that the systemd timer doesn't appear to be installed correctly.

[Upon installing the service](https://github.com/cloudflare/cloudflared/blob/10e4a2940cec351c6ab40b76420853ab3d347d45/cmd/cloudflared/linux_service.go#L240), `cloudflared-update.timer` is simply started, however it's necessary to **enable and start** to be persistent across boots.

Additionally, upon manually starting and enabling `cloudflared-update.timer`, my systemd showed the timer(`systemctl list-timers --all`) with no 'NEXT' run time, meaning it would never run anyway.

~~I believe this to be the result of [`OnUnitActiveSec` in the timer](https://github.com/cloudflare/cloudflared/blob/10e4a2940cec351c6ab40b76420853ab3d347d45/cmd/cloudflared/linux_service.go#L76), which '[Defines a timer relative to when the unit the timer unit is activating was last activated.](https://www.freedesktop.org/software/systemd/man/systemd.timer.html#Options)', however the unit (`cloudflared-update.service`) is never(and doesn't need to be) manually started, and so this time doesn't exist.~~

~~Using `OnActiveSec` instead should fix this part of the issue.~~

**EDIT:** This is due to `OnUnitActiveSec` as I mentioned before, however it was necessary for that to be there in order for the timer to repeat. Therefore, `OnCalendar` should be used for the timer instead.

Tugzrida:
Fixed by #159

PR #159, opened 2019-12-17
Fix timer scheduling for systemd update service (#133)

cilki:
The `cloudflared-update.timer` is now scheduled correctly instead of displaying "n/a" for the next runtime in `systemctl list-timers cloudflared-update.timer`.

It may also be desirable to `enable` this service as mentioned in #133, but I didn't want to make that decision.

### DIFF
--- a/cmd/cloudflared/linux_service.go
+++ b/cmd/cloudflared/linux_service.go
@@ -73,7 +73,7 @@ ExecStart=/bin/bash -c '{{ .Path }} update; code=$?; if [ $code -eq 64 ]; then s
 Description=Update Argo Tunnel
 
 [Timer]
-OnUnitActiveSec=1d
+OnCalendar=daily
 
 [Install]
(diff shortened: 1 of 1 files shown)


sssilver:
Thank you so much for the contribution, @cilki !

acmacalister:
Agreed. Thanks @cilki. What do y'all think about adding `Persistent=true` option to the timer in case it is missed for some reason?

https://wiki.archlinux.org/index.php/Systemd/Timers

cilki:
@acmacalister I'd probably leave `Persistent` out because it would only have a significant effect for systems that are routinely shutdown before the timer goes off (and thus the updater would never run at all).

For everyone else, `Persistent=true` would manifest as an additional update check on boot when the system has been powered down for some time. I think it's better to always have the updater run after a few hours rather than immediately at boot time.

Issue #160, opened 2019-12-18
The ability of allocating memory thresholds the tunnel throughput


rueian:
Hi, I am interested in using cloudflared tunnel in production. While trying to find the maximum throughput of a single tunnel, I notice that it will keep allocating huge amount of memory while serving requests and lead to the speed of allocating memory being a bottleneck of throughput.

There is a screenshot taken from my test cluster on GCP:
![截圖 2019-12-12 下午5 17 36](https://user-images.githubusercontent.com/2727535/70792777-eaee9d00-1dd4-11ea-819b-a25f49eb93ef.png)

There are 3 waves on the graph, from left to right. The only change I made to get these different waves is scaling the machine where the cloudflared running on:
`1st wave is generated from cloudflared running on n1-highcpu-4 machine`
`2nd wave is generated from cloudflared running on n1-highcpu-8 machine`
`3rd wave is generated from cloudflared running on n1-highcpu-16 machine`

The 1st wave shows that, when running on `n1-highcpu-4`, cloudflared keeps allocating `~1GB/s` memory to serve maximum `~2.5K/s` request with `~75ms` latency collected by my client tool.

The 2nd wave shows that, after switching to `n1-highcpu-8`, cloudflared keeps allocating `~2GB/s` memory to serve maximum `~4K/s` request with `~50ms` latency collected by my client tool.

The 3rd wave shows that, after switching to `n1-highcpu-16`, cloudflared keeps allocating `~2.8GB/s` memory to serve maximum `~5K/s` request with `~48ms` latency collected by my client tool.

However the top-left chart shows that the total memory keeps around 190MB.
So, It seems that cloudflared is quite busy on allocating memory and releasing memory. And it seems that if cloudflared can allocate memory faster, the latency will be slower and the throughput will be higher.

Luckily, after collecting the profiling graph, it does shows where is the cause.

<img width="1620" alt="截圖 2019-12-18 下午6 20 43" src="https://user-images.githubusercontent.com/2727535/71080161-9463d300-21c7-11ea-9aed-b8eb8e33324d.png">

It shows that almost all allocations come from the `TunnelHandler.serveHTTP`. I believed it is this line of code doing the allocation, and can be optimized with a buffer pool mechanism: 
https://github.com/cloudflare/cloudflared/blob/789ca6f6f46a72b5138a1700e476e8c1d04528bf/origin/tunnel.go#L645

rueian:
[...]

PR #161, opened 2019-12-18
perf(cloudflared): reuse memory from buffer pool to get better throughput

rueian:
There are three lines of code using `io.CopyBuffer()` with `make([]byte, 512*1024)` directly.
And they will make the cloudflared be very busy on allocating and releasing memory and will threshold the tunnel throughput by increasing latency.

This PR try to replace these `make([]byte, 512*1024)` with a `sync.Pool` to avoid allocating the buffer on every requests.

Ref: https://github.com/cloudflare/cloudflared/issues/160

### DIFF
--- a/buffer/pool.go
+++ b/buffer/pool.go
@@ -1,0 +1,27 @@
+package buffer
+
+import (
+	"sync"
+)
+
+type Pool struct {
+	buffers sync.Pool
+}
+
+func NewPool(bufferSize int) *Pool {
+	return &Pool{
+		buffers: sync.Pool{
+			New: func() interface{} {
+				return make([]byte, bufferSize)
+			},
+		},
+	}
+}
+
+func (p *Pool) Get() []byte {
+	return p.buffers.Get().([]byte)
+}
+
+func (p *Pool) Put(buf []byte) {
+	p.buffers.Put(buf)
+}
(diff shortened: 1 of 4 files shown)


buffer/pool.go
```diff
@@ -0,0 +1,27 @@
+package buffer
+
+import (
+	"sync"
+)
+
+type Pool struct {
+	buffers sync.Pool
```
nickvollmar: In the [docs for sync.Pool](https://golang.org/pkg/sync/#Pool) there is a note: "A Pool must not be copied after first use." I think we should also add "must not be copied" as a documentation item on this struct.

rueian: Thanks, I have added it in a new commit.

BTW, actually the `go vet` can detect the copy. Try `go vet` on this code:
```golang
package main

import (
	"github.com/cloudflare/cloudflared/buffer"
)

func main() {
	poolPtr := buffer.NewPool(512 * 1024)
	copy := *poolPtr

	copy.Get()
}
```
it will output:
```shell
▶ go vet main.go
# command-line-arguments
./main.go:9:10: assignment copies lock value to copy: github.com/cloudflare/cloudflared/buffer.Pool contains sync.Pool contains sync.noCopy
```

Ref: https://golang.org/issues/8005#issuecomment-190753527

rueian:
Hi, I just resolved the conflict, is there anything else I can help to get this merged?

Issue #113, opened 2019-06-28
dns.google is not compatible with "cloudflared"


bigdargon:
Hi,
I am using debian and ubuntu operating systems, and I run cloudflared to use dns-over-https.

Previously I used https://dns.google.com/experimental address, everything works normally. But I switched to using a new address according to google's notification, I received an error message:
- If I use the address "https://dns.google/dns-query": level=error msg="failed to connect to an HTTPS backend \"https://dns.google/dns-query\"" error="returned status code 415"
- If I use the address "https://dns.google/resolve": level=error msg="failed to connect to an HTTPS backend \"https://dns.google/resolve\"" error="returned status code 400"
- If I use the address "https://8.8.8.8/dns-query": level=error msg="failed to connect to an HTTPS backend \"https://8.8.8.8/dns-query\"" error="returned status code 404"

What do I have to do to fix this problem? Thank you!

bigdargon:
I created the issue here https://issuetracker.google.com/issues/136198937 and got the answer. Can Cloudflared update for compatibility?

> https://dns.google/dns-query is the right endpoint to use.
> 
> The problem is that (your version of) cloudflared is apparently still using the MIME type application/dns-udpwireformat rather than application/dns-message for the Content-Type header
> 
> From what I can tell, cloudflared is using the CoreDNS modular DNS server to implement its DNS to DoH proxy. The current version of CoreDNS DoH support (https://github.com/coredns/coredns/blob/master/plugin/pkg/doh/doh.go) uses application/dns-message, but the DoH support in CoreDNS was reconfigured and it is possible that cloudflared might need changes to work with the current CoreDNS architecture.
> 
> If you have been using the same version of cloudflared for a year or more, you should download a new copy and see if it still has this problem.
> 
> If the latest cloudflared still gives a 415 error, you might want to open an issue or request support on the Cloudflare community forum. If you are familiar with compiling Go programs, you could try to build cloudflared yourself and see what version of CoreDNS or the doh plugin it is using.
> 
> In the meantime, you can continue to use the https://dns.google.com/experimental endpoint, at least for another few weeks until you get the software issues sorted.

rezaxdi:
Ok, due to today's cloudlfare outage suddenly I was not able to visit any site because dns was down. I tried to use google dns in cloudflared client and it was not working. So I had to uninstall cloudflared and get back to dnscrypt. I think cloudflared needs to update it's client to be compatible with other DoH servers.

[...]

PR #168, opened 2020-01-25
Tweak HTTP host header. Fixes #107

### LINKED ISSUE #107: proxy-dns IPv6 upstream does not work (macOS 10.14.5)
Configuring proxy-dns as follows:

`$ ./cloudflared proxy-dns --metrics localhost:5555 --port 5054 --upstream "https://[2606:4700:4700::1111]/dns-query"`

starts up fine:

```
INFO[0000] Adding DNS upstream                           url="https://[2606:4700:4700::1111]/dns-query"
INFO[0000] Starting metrics server                       addr="127.0.0.1:5555"
INFO[0000] Starting DNS over HTTPS proxy server          addr="dns://localhost:5054"
```

but DNS queries fail with a 403:

```
ERRO[0011] failed to connect to an HTTPS backend "https://[2606:4700:4700::1111]/dns-query"  error="returned status code 403"
```

Looking at a pcap, it looks like the TLS connection is established OK, and that the dns-query request itself is denied by the server.

An example curl works as expected:

`$ curl -s -H 'accept: application/dns-message' 'https://[2606:4700:4700::1111]/dns-query?dns=q80BAAABAAAAAAAAA3d3dwdleGFtcGxlA2NvbQAAAQAB' | hexdump`

```0000000 ab cd 81 a0 00 01 00 01 00 00 00 01 03 77 77 77
0000010 07 65 78 61 6d 70 6c 65 03 63 6f 6d 00 00 01 00
0000020 01 c0 0c 00 01 00 01 00 00 1a 0d 00 04 5d b8 d8
0000030 22 00 00 29 05 ac 00 00 00 00 00 44 00 0c 00 40
0000040 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
*
0000080
```

Tugzrida:
I finally got the the root cause of the odd behaviour in #107.

It's to do with the HTTP Host headers accepted by 1.1.1.1.

With this working example, you can see that curl automatically sets the Host request header to `[2606:4700:4700::1111]`:

    $ echo -n 'q80BAAABAAAAAAAAA3d3dwdleGFtcGxlA2NvbQAAAQAB' | base64 -D | curl -v -H 'content-type: application/dns-message' --data-binary @- https://[2606:4700:4700::1111]/dns-query -o -
    ...
    > Host: [2606:4700:4700::1111]
    ...
    < HTTP/2 200
    ...

In this non-working example, the Host header is force-set to `2606:4700:4700::1111` (without brackets, as cloudflared does) and the server returns the 403 error seen when trying to use the IPv6 server in cloudflared:

    $ echo -n 'q80BAAABAAAAAAAAA3d3dwdleGFtcGxlA2NvbQAAAQAB' | base64 -D | curl -v -H 'content-type: application/dns-message' --data-binary @- https://[2606:4700:4700::1111]/dns-query -o - -H 'Host: 2606:4700:4700::1111'
    ...
    > Host: 2606:4700:4700::1111
    ...
    < HTTP/2 403
    ...

The simple way I found to fix this was using `u.endpoint.Host` instead of `u.endpoint.Hostname()` to set the Host header, which doesn't strip out the brackets. I'm new to Go, so I don't know if that'll have any side effects, but at least the root cause is now known and some other solution can be found.

### DIFF
--- a/tunneldns/https_upstream.go
+++ b/tunneldns/https_upstream.go
@@ -83,7 +83,7 @@ func (u *UpstreamHTTPS) exchangeWireformat(msg []byte) ([]byte, error) {
 	}
 
 	req.Header.Add("Content-Type", "application/dns-message")
-	req.Host = u.endpoint.Hostname()
+	req.Host = u.endpoint.Host
 
 	resp, err := u.client.Do(req)
 	if err != nil {


einichi:
Great work, I was looking into this as I have also been trying to configure IPv6 upstream servers and was experiencing the same issue. Saved me some time!

So either merge this or commit something equivalent, or change the server-side behavior. Based on behavior I've observed in other applications however, stripping the brackets is incorrect, so I believe this should be patched in the client as you have done.

@adamchalmers What do you think?

Tugzrida:
According to the [RFC for the host header](https://tools.ietf.org/html/rfc7230#section-5.4) and [the definition of `uri-host`](https://tools.ietf.org/html/rfc3986#section-3.2.2):

> The host subcomponent of authority is identified by an IP literal encapsulated within square brackets, an IPv4 address in dotted-decimal form, or a registered name.

which matches the behaviour in every http client I've observed.

According to the go docs, [`URL.Host` includes the host and port](https://golang.org/pkg/net/url/#URL), whereas [`URL.Hostname()` specifically strips the port and square brackets](https://golang.org/pkg/net/url/#URL.Hostname). Comparing the RFCs with these, `URL.Host` is the right way to go.

Something else we might want to consider is allowing a user-override to the Host header as some DoH providers do not accept IP literals at all, and require a hostname, which of course can cause bootstrapping problems if used directly in the upstream URL as would currently be necessary. This has come up in #113 with `dns.google`.

PR #171, opened 2020-02-09
Fix typo in debug logging level warning

ryanhofer:

PR #174, opened 2020-02-20
Assorted typo fixes

XhmikosR:
BTW I left the vendor folder untouched since I assume that's third-party code.


XhmikosR:
@adamchalmers you should also merge #171; I left it out since there was a PR open :)

XhmikosR:
Friendly ping about this and #171. I know these are not important fixes, but it's better than having PRs hanging for ever :)

PR #181, opened 2020-03-11
proxy-dns: Allow setting the Host header via url hash

### LINKED ISSUE #113: dns.google is not compatible with "cloudflared"

Tugzrida:
Some DoH providers (namely Google) won't accept IP addresses in the Host header of DoH requests.

As I mentioned in #113, using dns.google with cloudflared currently requires using the url `https://dns.google/dns-query`. This however causes a bootstrapping issue if cloudflared has been set as the system's only resolver, as dns.google can't be resolved. dns.google therefore needs to be added to the system's host file. Even then, this solution is not ideal, as the host file can only define one IP address per hostname.

This PR allows the HTTP Host header and TLS verification name to be overridden by adding a hash/anchor/fragment to the end of the URL. In normal HTTP implementations, the segment of the URL after the hash is not sent to the server, and in DoH specifically, it serves no purpose.

After the implementation of this change, services that require a specific Host header like dns.google could be used with upstream settings like this:

```
proxy-dns-upstream:
 - https://8.8.8.8/dns-query#dns.google
 - https://8.8.4.4/dns-query#dns.google
 - https://[2001:4860:4860::8888]/dns-query#dns.google
 - https://[2001:4860:4860::8844]/dns-query#dns.google
```

I also added Cloudflare DNS's IPv6 addresses to the default DoH configuration while I was editing the argument list, as IPv6 support was fixed in #168.

Fixes #113

### DIFF
--- a/cmd/cloudflared/tunnel/cmd.go
+++ b/cmd/cloudflared/tunnel/cmd.go
@@ -129,8 +129,8 @@ func Commands() []*cli.Command {
 				},
 				&cli.StringSliceFlag{
 					Name:    "upstream",
-					Usage:   "Upstream endpoint URL, you can specify multiple endpoints for redundancy.",
-					Value:   cli.NewStringSlice("https://1.1.1.1/dns-query", "https://1.0.0.1/dns-query"),
+					Usage:   "Upstream endpoint URL, you can specify multiple endpoints for redundancy. If required, the Host header can be manually set by appending a hash to the URL.",
+					Value:   cli.NewStringSlice("https://1.1.1.1/dns-query", "https://[2606:4700:4700::1111]/dns-query", "https://1.0.0.1/dns-query", "https://[2606:4700:4700::1001]/dns-query"),
 					EnvVars: []string{"TUNNEL_DNS_UPSTREAM"},
 				},
 			},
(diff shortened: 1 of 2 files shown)


Tugzrida:
I was just playing around with Cloudflare Gateway and discovered that this is an issue there too. 

Setting cloudflared to connect to `https://random-id.cloudflare-gateway.com/dns-query` doesn't work as the machine cannot bootstrap resolve that domain.

With this PR, entering `https://162.159.36.5/dns-query#random-id.cloudflare-gateway.com` would allow Cloudflare Gateway to work with cloudflared.

Looking at that, it may be more neat to put the host and IP the other way around: `https://random-id.cloudflare-gateway.com/dns-query#162.159.36.5`

I'm happy to work on either way if there's any suggestions.

Tugzrida:
Closing this as it was fixed far more comprehensively in f18209a

Issue #59, opened 2018-11-19
Multiple tunnels


mnaser:
Hi there,

So, we're slowly becoming heavy users of Argo Tunnel, especially with the SSH features that are coming up and we're slowly realizing there are use cases where we need to run multiple tunnels.

The current configuration assumes that 1 config file contains 1 tunnel only, which can be problematic because at times we want to expose the both the service (HTTP) and SSH via CloudFlare.

There's two ways that we can really do this right now:

*systemd templates*
This is the way that we're currently working with, it would involve a change here:

https://github.com/cloudflare/cloudflared/blob/41916365b689bf2cc1446ea5717e4d26cc8aed43/cmd/cloudflared/linux_service.go#L43-L59

Once that section is updated, we can use templated systemd files: https://fedoramagazine.org/systemd-template-unit-files/

This means that you can do something like `systemctl start cloudflared@ssh` and it will start it up by using the config file in `/etc/cloudflared/ssh.yml` -- this means that all you need to do is drop a file in `/etc/cloudflared` and then start up that unit.

*config file change*
This one is a bit more complex and it might be not backwards compatible but simply having the same executable create multiple tunnels might be super productive and useful.  It would obviously imply a lot more work (but I don't know the internal architecture and how easy that might be).

*Use-case*
I know this is really important especially with pushing Argo as a VPN replacement.  In our case, we want a "host" that pretty much does tunnelling for a lot of devices that can't run cloudflared.  For example, iDRAC for servers, SSH for switches, etc.

The second solution is obviously way better in terms of scale and memory footprint, especially if you're in an environment where possibly you can be doing this for 200-300 hosts, 200-300 services might be really difficult to manage and a single configuration file might be much easier.

tanner-bruce:
> especially with the SSH features that are coming up

I've been searching around for this and can't find any info. Do you have a link where I could read about this? If it is private I can ask my account rep. Thanks

mnaser:
Hey @tanner-bruce 

It's no secret, this is the blog post at CloudFlare:

https://blog.cloudflare.com/releasing-the-cloudflare-access-feature-that-let-us-smash-a-vpn-on-stage/

mnaser:
Also, we've worked around it with the following templates that has been working great for us (named `cloudflared@.service`)

~~~~
[Unit]
Description=Argo Tunnel (%I)
After=network.target

[Service]
TimeoutStartSec=0
Type=notify
[...]

PR #197, opened 2020-04-29
added option to map a cname host to a local origin

udnaan:
This add support for multiple local origin through a single cloudflared process.
Config file example to use it is as follows:

```
hostname: public.opus.ai
 # default url that will use any requests not coming through cname hosts
url: http://192.168.3.10:8080
logfile: ./cloudflared.log
origincert: ./cert.pem
no-autoupdate: true
host-to-origin:
    # first service cnamed to public.example.com
    - service1.opus.ai=http://localhost:8080
    # second service cnamed to public.example.com
    - service2.opus.ai=http://192.168.3.10:8080
   .... # and so on
```

Note that each service still needs to be cnamed manually in cloudflare panel. It should be possible to use the api through the service itself but I haven't explored it yet.

### DIFF
--- a/cmd/cloudflared/tunnel/cmd.go
+++ b/cmd/cloudflared/tunnel/cmd.go
@@ -768,6 +768,12 @@ func tunnelFlags(shouldHide bool) []cli.Flag {
 			EnvVars: []string{"TUNNEL_HOSTNAME"},
 			Hidden:  shouldHide,
 		}),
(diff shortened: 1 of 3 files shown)


udnaan:
Addresses https://github.com/cloudflare/cloudflared/issues/59

sudarshan-reddy:
Closing thus because this now exists as ingress rules.

PR #218, opened 2020-07-06
beautify package meta information generated by fpm

hoexter:
Current Debian and RPM meta information are very generic, e.g.
$ apt-cache show cloudflared
Package: cloudflared
Version: 2020.7.0
License: unknown
Vendor: @k8s-managed-krwtk.teamcity-agents.svc.cluster.local
Architecture: amd64
Maintainer: <@k8s-managed-krwtk.teamcity-agents.svc.cluster.local>
Installed-Size: 42883
Section: default
Priority: extra
Homepage: http://example.com/no-uri-given
Description: no description given

Move those generic settings to `FPM_COMMONG_FLAGS` and pass those
to fpm when building the package.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -92,6 +92,11 @@ define build_package
 	cp cloudflared $(PACKAGE_DIR)/cloudflared
 	cat cloudflared_man_template | sed -e 's/\$${VERSION}/$(VERSION)/; s/\$${DATE}/$(DATE)/' > $(PACKAGE_DIR)/cloudflared.1
 	fakeroot fpm -C $(PACKAGE_DIR) -s dir -t $(1) --$(1)-compression bzip2 \
+		--description 'Cloudflare Argo tunnel daemon' \
+		--vendor 'Cloudflare' \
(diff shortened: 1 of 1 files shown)


hoexter:
I'm not sure which fpm release you use internally to build, but according to https://github.com/jordansissel/fpm/issues/1685 and my own test fpm creates broken packages for deb compression types != gz. You end up with a control.tar.gz file either compressed by bzip2 or xz, which dpkg is not able to deal with. So I presonally would drop the `--deb-compression bzip` to be on the safe side. But since the package in your repo is ok, I believe you either fixed it locally or use an older release.

Michael9127:
@hoexter Thanks for opening up this PR! the deb-compression issue is a great thing to note. Yes, we are most likely using an older version of FPM. Looks like there is a merge conflict, but other than that it looks good to me.

hoexter:
Oh well seems in the meantime a similar cleanup got merged from a few days before. I see what I can do with just moving some of the flags for the meta information.

PR #220, opened 2020-07-07
drop usage of cat when sed is already invoked

hoexter:
ymmv :)


sudarshan-reddy:
@hoexter sorry for the late response. This seems like an acceptable change. Would you like to rebase so we can accept this PR?

sudarshan-reddy (approved):

sudarshan-reddy:
@sssilver , @joliveirinha can we merge this?

PR #238, opened 2020-09-17
Updater service exit code should be 11

BojanZelic:
Updater service should have exit code of 11 and not 64;

See: https://github.com/cloudflare/cloudflared/blob/a7562dff68d92210e9bb774916e5fe9198724918/cmd/cloudflared/updater/update.go#L51

### DIFF
--- a/cmd/cloudflared/linux_service.go
+++ b/cmd/cloudflared/linux_service.go
@@ -67,7 +67,7 @@ Description=Update Argo Tunnel
 After=network.target
 
 [Service]
-ExecStart=/bin/bash -c '{{ .Path }} update; code=$?; if [ $code -eq 64 ]; then systemctl restart cloudflared; exit 0; fi; exit $code'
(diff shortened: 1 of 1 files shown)


sssilver:
Hi @BojanZelic -- thanks for your contribution!

Can you elaborate on why 11 is better than 64 here?

adamchalmers:
@sssilver because [that's what cloudflared exits with when it autoupdates](https://github.com/cloudflare/cloudflared/blob/a7562dff68d92210e9bb774916e5fe9198724918/cmd/cloudflared/updater/update.go#L51) from the changes @chungthuang made to resolve the last autoupdate bug

adamchalmers (approved):

BojanZelic:
Currently the cloudflared-update service fails to restart the cloudflared service after updating the bin. 

This is because it checks for exit code 64 instead of 11 (11 is the exit code that the updater exists with when it updates successfully)

chungthuang (approved):

sssilver (approved):

PR #263, opened 2020-11-18
Update error message to use login command

securitygeneration:
Unless I'm mistaken, when there is no existing token for an app, the `login` command needs to be run to obtain a token (not the `token` command, which itself doesn't generate a token).


adamchalmers (approved):

daltoniam (approved):

sssilver (approved):

adamchalmers:
Thanks for the PR! Sorry we didn't see it for so long heh

PR #267, opened 2020-11-19
feat(tunnel): add TUNNEL_ORIGIN_CERT_CONTENT env

pims:
This PR adds support for specifying the origin certificate content
as an environment variable.
This is useful when deploying cloudflared in environments where writing
a file is not possible, ex: Heroku

The cert must be base64 encoded.

### DIFF
--- a/cmd/cloudflared/tunnel/configuration.go
+++ b/cmd/cloudflared/tunnel/configuration.go
@@ -2,6 +2,7 @@ package tunnel
 
 import (
 	"crypto/tls"
+	"encoding/base64"
 	"fmt"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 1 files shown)


sodabrew:
Drive-by chime in: I used https://github.com/hashicorp/consul-template with the pem file defined by a template like
```
{{ env "TUNNEL_ORIGIN_CERT_CONTENT" }}
```
which aligned with other Consul discovery lookups I was doing, but I would agree with you if you argued that it's overkill for just the use case of taking an injected environment variable and dropping it into a file. +1 for this feature, I would use it!

adamchalmers (approved):

AiScripter:
This feature is very useful, is there a reason this pull request has been closed ?

AiScripter:
@sudarshan-reddy  (sorry for the mention)

Issue #91, opened 2019-04-30
error msg="failed to connect to an HTTPS backend \"https://1.1.1.1/dns-query\"" error="failed to perform an HTTPS request: Post https://1.1.1.1/dns-query: net/http: request canceled (Client.Timeout exceeded while awaiting headers)"


mflannery:
I have configured my home DNS to use dnsmasq and cloudflared on a Fedora 29 system. If I use https://1.1.1.1/dns-query as the upsteam, I see the following errors happen occasionally:

Apr 30 15:24:12 dns-server cloudflared[20706]: time="2019-04-30T15:24:12-04:00" level=error msg="failed to connect to an HTTPS backend \"https://1.1.1.1/dns-query\"" error="failed to perform an HTTPS request: Post https://1.1.1.1/dns-query: net/http: request canceled (Client.Timeout exceeded while awaiting headers)"

But DNS lookups seem to work all the time, even with the error.

If I change my upstream to https://dns.adguard.com/dns-query (AD and malicious site blocking at DNS level) then I see lots of these errors in the logs:

Apr 30 14:52:20 dns-server cloudflared[20680]: time="2019-04-30T14:52:20-04:00" level=error msg="failed to connect to an HTTPS backend \"https://dns.adguard.com/dns-query\"" error="failed to perform an HTTPS request: Post https://dns.adguard.com/dns-query: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)"

And DNS lookups might work once or twice but then stop working all together.

Any ideas what's going on here?  I love the security of DoH and ad blocking and would prefer not to have to maintain the ad blocking myself (even though it is admittedly not difficult).

Thanks!
Mike

mflannery:
I have confirmed this error also occurs with https://doh.cleanbrowsing.org/doh/security-filter/ as the upstream.

To install, I followed this guide: https://bendews.com/posts/implement-dns-over-https/ except that instead of putting a config file in /etc/default, I used /etc/cloudflared/config.yml.  My config.yml file looks like this:

proxy-dns: true
proxy-dns-upstream:
#- https://dns.adguard.com/dns-query
- https://1.1.1.1/dns-query
- https://1.0.0.1/dns-query
proxy-dns-port: 5053

and my systemd service file looks like this:

[Unit]
Description=cloudflared DNS over HTTPS proxy
After=syslog.target network-online.target

[Service]
Type=simple
User=cloudflared
[...]

PR #290, opened 2021-01-09
Add max upstream connections dns-proxy option

dvejmz:
Allows defining a limit to the number of connections that can be established with the upstream DNS host. If left unset, there may be situations where connections fail to establish, which causes the Transport to create an influx of connections causing upstream to throttle our requests and triggering a runaway effect resulting in high CPU usage.

This is a workaround for [the unresolved upstream connection issues](https://github.com/cloudflare/cloudflared/issues/23#issuecomment-687743790) that some people have reported occurring in Raspberry Pis running Pi-hole. By allowing users to cap the number of connections themselves to adjust them to the capacity of their specific device, we're at least mitigating against the impact this runaway effect has on the limited CPU resources these devices have.

```
Jan 09 17:38:46 hostname cloudflared[9896]: 5:38PM ERR failed to connect to an HTTPS backend "https://1.0.0.1/dns-query" error="failed to perform an HTTPS request: Post \"https://1.0.0.1/dns-query\": context deadline exceeded"
```

I'm currently running a cloudflared binary compiled with this changeset like so

```
cloudflared proxy-dns --port 54 --upstream https://1.1.1.1/dns-query --upstream https://1.0.0.1/dns-query --max-upstream-conns 5
```

Which brings down the %CPU of the cloudflared task in my Raspberry Pi 1 Model B from 80-98% to the 0-5% range.

This fix is based on @kitsook's [changeset](https://github.com/kitsook/cloudflared/commit/134410ae07e559743d50ee516b8b4a5111e01200). See #91 for more details about the issue.

### DIFF
--- a/cmd/cloudflared/config/model.go
+++ b/cmd/cloudflared/config/model.go
@@ -30,6 +30,7 @@ type DNSResolver struct {
 	Port       uint16   `json:"port,omitempty"`
 	Upstreams  []string `json:"upstreams,omitempty"`
 	Bootstraps []string `json:"bootstraps,omitempty"`
+	MaxUpstreamConnections int `json:"max_upstream_connections,omitempty"`
 }
 
 // Root is the base options to configure the service
@@ -59,6 +60,7 @@ func (r *DNSResolver) Hash() string {
 	io.WriteString(h, strings.Join(r.Bootstraps, ","))
 	io.WriteString(h, strings.Join(r.Upstreams, ","))
 	io.WriteString(h, fmt.Sprintf("%d", r.Port))
+	io.WriteString(h, fmt.Sprintf("%d", r.MaxUpstreamConnections))
 	io.WriteString(h, fmt.Sprintf("%v", r.Enabled))
 	return fmt.Sprintf("%x", h.Sum(nil))
 }
@@ -98,4 +100,12 @@ func (r *DNSResolver) BootstrapsOrDefault() []string {
 		return r.Bootstraps
 	}
 	return []string{"https://162.159.36.1/dns-query", "https://162.159.46.1/dns-query", "https://[2606:4700:4700::1111]/dns-query", "https://[2606:4700:4700::1001]/dns-query"}
+}
+
+// MaxUpstreamConnectionsOrDefault return the max upstream connections or returns the default if 0
+func (r *DNSResolver) MaxUpstreamConnectionsOrDefault() int {
+    if r.MaxUpstreamConnections >= 0 {
+        return r.MaxUpstreamConnections
+    }
+    return 0
 }
--- a/cmd/cloudflared/config/model.go
+++ b/cmd/cloudflared/config/model.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"io"
 	"strings"
+
+	"github.com/cloudflare/cloudflared/tunneldns"
 )
 
 // Forwarder represents a client side listener to forward traffic to the edge
@@ -25,12 +27,12 @@ type Tunnel struct {
 
 // DNSResolver represents a client side DNS resolver
 type DNSResolver struct {
-	Enabled    bool     `json:"enabled"`
-	Address    string   `json:"address,omitempty"`
-	Port       uint16   `json:"port,omitempty"`
-	Upstreams  []string `json:"upstreams,omitempty"`
-	Bootstraps []string `json:"bootstraps,omitempty"`
-	MaxUpstreamConnections int `json:"max_upstream_connections,omitempty"`
+	Enabled                bool     `json:"enabled"`
+	Address                string   `json:"address,omitempty"`
+	Port                   uint16   `json:"port,omitempty"`
+	Upstreams              []string `json:"upstreams,omitempty"`
+	Bootstraps             []string `json:"bootstraps,omitempty"`
+	MaxUpstreamConnections int      `json:"max_upstream_connections,omitempty"`
 }
 
 // Root is the base options to configure the service
@@ -102,10 +104,10 @@ func (r *DNSResolver) BootstrapsOrDefault() []string {
 	return []string{"https://162.159.36.1/dns-query", "https://162.159.46.1/dns-query", "https://[2606:4700:4700::1111]/dns-query", "https://[2606:4700:4700::1001]/dns-query"}
 }
 
-// MaxUpstreamConnectionsOrDefault return the max upstream connections or returns the default if 0
+// MaxUpstreamConnectionsOrDefault return the max upstream connections or returns the default if negative
 func (r *DNSResolver) MaxUpstreamConnectionsOrDefault() int {
-    if r.MaxUpstreamConnections >= 0 {
-        return r.MaxUpstreamConnections
-    }
-    return 0
+	if r.MaxUpstreamConnections >= 0 {
+		return r.MaxUpstreamConnections
+	}
+	return tunneldns.MaxUpstreamConnsDefault
 }
--- a/cmd/cloudflared/tunnel/cmd.go
+++ b/cmd/cloudflared/tunnel/cmd.go
@@ -1004,7 +1004,7 @@ func configureProxyDNSFlags(shouldHide bool) []cli.Flag {
 			Hidden:  shouldHide,
 		}),
 		altsrc.NewIntFlag(&cli.IntFlag{
-			Name:    tunneldns.MaxUpstreamConnsFlag,
+			Name:    "proxy-dns-max-upstream-conns",
 			Usage:   "Maximum concurrent connections to upstream. Setting to 0 means unlimited.",
 			Value:   tunneldns.MaxUpstreamConnsDefault,
 			Hidden:  shouldHide,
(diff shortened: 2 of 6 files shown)


adamchalmers:
Great, thanks very much for this PR! @sudarshan-reddy reproduced and confirmed this fixes it.

cmd/cloudflared/tunnel/server.go
```diff
@@ -13,7 +13,11 @@ func runDNSProxyServer(c *cli.Context, dnsReadySignal, shutdownC chan struct{},
 	if port <= 0 || port > 65535 {
 		return errors.New("The 'proxy-dns-port' must be a valid port number in <1, 65535> range.")
 	}
-	listener, err := tunneldns.CreateListener(c.String("proxy-dns-address"), uint16(port), c.StringSlice("proxy-dns-upstream"), c.StringSlice("proxy-dns-bootstrap"), log)
+	maxUpstreamConnections := c.Int("proxy-dns-max-upstream-conns")
```
adamchalmers, line 16: Shouldn't this be `c.Int("max-upstream-conns")` to match the [definition](https://github.com/cloudflare/cloudflared/pull/290/files?diff=unified&w=1#diff-6f1cb455303a0da3ac60ed8b8cdcb845a5edb82939f56e2e05939cb625522a76)

adamchalmers (changes requested):

tunneldns/tunnel.go
```diff
@@ -68,6 +68,12 @@ func Command(hidden bool) *cli.Command {
 				Value:   cli.NewStringSlice("https://162.159.36.1/dns-query", "https://162.159.46.1/dns-query", "https://[2606:4700:4700::1111]/dns-query", "https://[2606:4700:4700::1001]/dns-query"),
 				EnvVars: []string{"TUNNEL_DNS_BOOTSTRAP"},
 			},
+			&cli.IntFlag{
+				Name:    "max-upstream-conns",
+				Usage:   "Maximum concurrent connections to upstream, unlimited by default",
+				Value:   0,
```
adamchalmers, line 74: I think 0 is a dangerous default here, because it'll cause the bad behaviour. Let's make the default a positive number. I'm not sure what is appropriate here... 10? Maybe 100? And let's put it into a `const` so that other code can reference the default too (e.g. `func configureBootstrap` and `func MaxUpstreamConnectionsOrDefault`).

adamchalmers (changes requested):

adamchalmers:
@dvejmz @sudarshan-reddy I pushed up a new commit, could you let me know what you think?

cmd/cloudflared/config/model.go
```diff
@@ -25,11 +27,12 @@ type Tunnel struct {
...
-	Address    string   `json:"address,omitempty"`
-	Port       uint16   `json:"port,omitempty"`
-	Upstreams  []string `json:"upstreams,omitempty"`
-	Bootstraps []string `json:"bootstraps,omitempty"`
+	Enabled                bool     `json:"enabled"`
+	Address                string   `json:"address,omitempty"`
+	Port                   uint16   `json:"port,omitempty"`
+	Upstreams              []string `json:"upstreams,omitempty"`
+	Bootstraps             []string `json:"bootstraps,omitempty"`
+	MaxUpstreamConnections int      `json:"max_upstream_connections,omitempty"`
```
sudarshan-reddy, line 35: I wonder if this setting belongs here: https://github.com/cloudflare/cloudflared/blob/master/cmd/cloudflared/tunnel/cmd.go#L985 instead.

sudarshan-reddy (commented):
The revision LGTM. I just have one question.

cmd/cloudflared/config/model.go
ipostelnik: Do we really need this to be configurable? How would the users know how to set this? Can we pick a reasonable default and go with it? Say 100 connections or even 1000?

sudarshan-reddy: I like your point. We could set that default here: https://github.com/cloudflare/cloudflared/pull/290/commits/c2f51021199b57f7d7f104526208b52763dc6984#diff-6f1cb455303a0da3ac60ed8b8cdcb845a5edb82939f56e2e05939cb625522a76R27 and still keep it override-able if anyone should choose to want a different number (even 0).

adamchalmers: We should probably never let the user choose 0 as it will crash their machine by eating 100% CPU, so yeah, maybe Igor's right and this doesn't need to be configurable.

kitsook: I think having it configurable makes sense. Say Raspberry Pi may see 100% CPU with 10 concurrent connections. But an enterprise server may easily handle 50 connections.

If there is concern with unlimited upper limit, maybe set the min as 1 and disallow setting it to 0?

As for the default value, it should be something that won't trigger throttling with most DoH providers. Say if the provider has a limit of 5 and the default max is 10, in the worst case, the provide may just reject requests from all 10 connections while cloudflared keeps re-creating 10 connections in the pool.

dvejmz: Thanks all for your comments on this. I will review your suggestions this evening and update the PR. I do agree providing a default value would save the user the trial-and-error process involved in tweaking it to their hardware. For me, a value of 5 upstream connections gives me really good results on a RPi 1 Model B, but it may be possible to increase it if necessary on multi-core devices.

adamchalmers: Default of 5 sounds good, at least we know it works for some users. And let's keep it configurable, it should be able to run well on any hardware.

dvejmz:
I have pushed some changes addressing your feedback:

- I've appended the `max-upstream-conns` CLI flag to `tunneldns/tunnel.go` as per @sudarshan-reddy 's suggestion. I have not removed it from `cmd/cloudflared/config/model.go` as I think it's still required there so that we can initialise the `tunneldns` listener within the `ResolverService` wrapper. Please let me know if there's something I misunderstood in your feedback, I'll be happy to change it.
- I've reduced the default max upstream connections from `10` to `5` @adamchalmers . I've not been able to reproduce the issue today on my Raspberry Pi so I didn't get a chance to verify if `10` would actually be a safe default as well. I know a value of `5` is guaranteed to work as I did confirm it performed correctly when connections to the HTTPS upstreams fail. Further testing in the future may lead to it being increased to 10 (or higher) in the future. I suppose as long as it remains configurable, it shouldn't be an issue either way.

I hope that makes sense. I recompiled and reinstalled this newer patched version in my RPi and everything seems to be working nicely. Do let me know if you have any other comments!

adamchalmers (approved):

cmd/cloudflared/tunnel/cmd.go
```diff
@@ -1003,6 +1003,13 @@ func configureProxyDNSFlags(shouldHide bool) []cli.Flag {
 			EnvVars: []string{"TUNNEL_DNS_UPSTREAM"},
 			Hidden:  shouldHide,
 		}),
+		altsrc.NewIntFlag(&cli.IntFlag{
+			Name:    tunneldns.MaxUpstreamConnsFlag,
```
ipostelnik, line 1007: You can't use `max-upstream-conns` as flag name here, it's not clear that it's used for tunneling DNS only. You need to prefix it with `proxy-dns-`. You'll need to change the code that reads it as well.

dvejmz: Good spot, will fix, thanks!

ipostelnik (approved):

ipostelnik:
@dvejmz thanks for implementing these changes

Stephen988:
Hi @dvejmz can you explain step by step how to fix cloudflared? I don't understand where apply this fix for reduce max connections.
Thanks!

dvejmz:
HI @Stephen988 ! In order to apply this fix, make sure you've downloaded the latest cloudflared binary (or at least any version equal or later than [2021.2.3](https://github.com/cloudflare/cloudflared/releases/tag/2021.2.3)). If you cannot find the right binary for your machine/architecture ─ if you're running it in an older Raspberry Pi like a Pi 1 ─ you may need to compile `cloudflared` yourself. To do that, make sure you've got `go` installed and then build it from the git repo with a command similar to the one below, substituting your target OS and architecture in the env vars below. The build config below will build it for a Raspberry Pi 1.

```
GOOS=linux GOARCH=arm GOARM=6 go build -v "-ldflags=-X 'main.Version=dev' -X 'main.BuildTime=$(date)'" ./cmd/cloudflared
```

Once you've got the cloudflared binary in the Pi, you can launch it as normal, as it should work fine with the current default max connections parameter. If you wish to change this parameter regardless, you'll need to pass it as an option when launching `cloudflared` as below

```
cloudflared proxy-dns [other flags you may be already using here] --max-upstream-conns <desired_number>
```

Hope that helps.

Stephen988:
Hi @dvejmz I'm very noob, never compiled a software, I have a raspberry pi 4 running raspbian os 32bit, actually I have the most recent version of docker, portainer and I'm using crazymax/cloudflared:latest.
how and where insert your string? sorry but is a new field for me.

kitsook:
@Stephen988 You can set the parameter via environment variable, e.g.
```
docker run ... --env TUNNEL_DNS_MAX_UPSTREAM_CONNS=<desired_number> ...
```

Stephen988:
> @Stephen988 You can set the parameter via environment variable, e.g.
> 
> ```
> docker run ... --env TUNNEL_DNS_MAX_UPSTREAM_CONNS=<desired_number> ...
> ```

nothing changes, always error: ERR failed to connect to an HTTPS backend "https://1.1.1.1/dns-query" error="failed to perform an HTTPS request: Post \"https://1.1.1.1/dns-query\": context deadline exceeded (Client.Timeout exceeded while awaiting headers)"

PR #330, opened 2021-03-08
Allow partial reads from a GorillaConn; add SetDeadline (from net.Conn)

benbuzbee:
The current implementation of GorillaConn will drop data if the
websocket frame isn't read 100%. For example, if a websocket frame is
size=3, and Read() is called with a []byte of len=1, the 2 other bytes
in the frame are lost forever.

This is currently masked by the fact that this is used primarily in
io.Copy to another socket (in ingress.Stream) - as long as the read buffer
used by io.Copy is big enough (it is 32*1024, so in theory we could see
this today?) then data is copied over to the other socket.

The client then can do partial reads just fine as the kernel will take
care of the buffer from here on out.

I hit this by trying to create my own tunnel and avoiding
ingress.Stream, but this could be a real bug today I think if a
websocket frame bigger than 32*1024 was received, although it is also
possible that we are lucky and the upstream size which I haven't checked
uses a smaller buffer than that always.

The test I added hangs before my change, succeeds after.

Also add SetDeadline so that GorillaConn fully implements net.Conn

### DIFF
--- a/websocket/connection.go
+++ b/websocket/connection.go
@@ -44,8 +44,15 @@ func (c *GorillaConn) Read(p []byte) (int, error) {
 	if err != nil {
 		return 0, err
 	}
-	// Write into an intermediate buffer in case the websocket frame contains more data then len(p)
-	// Write returns nil error always and grows the buffer everything is always written or panic
+
+	// Fast path, no need to stage in the intermediate buffer
+	if len(p) >= len(message) {
+		return copy(p, message), nil
+	}
+
+	// Slow path when the entire frame won't fit in p, stage through a readBuffer so the unread bytes
+	// aren't lost
+	// Write returns a nil error always and grows the buffer; everything is always written or panic
 	c.readBuf.Write(message)
 	return c.readBuf.Read(p)
 }
--- a/websocket/connection.go
+++ b/websocket/connection.go
@@ -1,7 +1,9 @@
 package websocket
 
 import (
+	"bytes"
 	"context"
+	"fmt"
 	"io"
 	"time"
 
@@ -27,17 +29,25 @@ const (
 // This is still used by access carrier
 type GorillaConn struct {
 	*websocket.Conn
-	log *zerolog.Logger
+	log     *zerolog.Logger
+	readBuf bytes.Buffer
 }
 
 // Read will read messages from the websocket connection
 func (c *GorillaConn) Read(p []byte) (int, error) {
+	// Intermediate buffer may contain unread bytes from the last read, start there before blocking on a new frame
+	if c.readBuf.Len() > 0 {
+		return c.readBuf.Read(p)
+	}
+
 	_, message, err := c.Conn.ReadMessage()
 	if err != nil {
 		return 0, err
 	}
-	return copy(p, message), nil
-
+	// Write into an intermediate buffer in case the websocket frame contains more data then len(p)
+	// Write returns nil error always and grows the buffer everything is always written or panic
+	c.readBuf.Write(message)
+	return c.readBuf.Read(p)
 }
 
 // Write will write messages to the websocket connection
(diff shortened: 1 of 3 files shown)


adamchalmers:
Thanks for the PR! I've added some team members to the reviewers. Quick question: you said you hit this by "avoiding ingress.Stream". How were you avoiding ingress.Stream?

benbuzbee:
Adam,
I'm using it like an SDK to avoid running cloudflared and also to avoid creating a tcp listener I don't need e.g.
ws, res, err := cfwebsocket.ClientConnect(req, websocket.DefaultDialer)
netCon := &GorillaConn{Conn: conn}
transport := &http.Transport{
  TLSHandshakeTimeout: time.Minute,
  DialContext := function that always returns netCon
}

This results in a TLS timeout in 1 minute as the go standard tls library reads with a small enough buffer that data gets lost, and it keeps waiting to read what it expects to read to complete the tls handshake.

ACK this probably isnt a supported scenario but thought you might want the fix

websocket/connection.go
```diff
@@ -27,17 +29,25 @@ const (
...
 	_, message, err := c.Conn.ReadMessage()
 	if err != nil {
 		return 0, err
 	}
-	return copy(p, message), nil
-
+	// Write into an intermediate buffer in case the websocket frame contains more data then len(p)
+	// Write returns nil error always and grows the buffer everything is always written or panic
+	c.readBuf.Write(message)
+	return c.readBuf.Read(p)
```
nmldiegues, line 50: Can we avoid the double copy in case we know it is not needed? (i.e., if `len(message) <= len(p)`)

websocket/connection.go
```diff
@@ -27,17 +29,25 @@ const (
...
+	}
+
 	_, message, err := c.Conn.ReadMessage()
 	if err != nil {
 		return 0, err
 	}
-	return copy(p, message), nil
-
+	// Write into an intermediate buffer in case the websocket frame contains more data then len(p)
+	// Write returns nil error always and grows the buffer everything is always written or panic
```
nmldiegues, line 48: This line seems to contain 2 sentences that could be split (maybe with a `;` after `buffer`?)

nmldiegues (commented):
LGTM.

Just had 1 suggestion

adamchalmers (approved):

websocket/connection.go
benbuzbee: just a philosphy I have to not fork the code unless performance is critical in order to not create a 'rarely used' path where bugs can develop unnoticed

Just sharing 'why', but I made the change :) there is now a fast path

mahdiamer (approved):

mahdiamer:
Oki

nmldiegues (approved):

websocket/connection.go
```diff
@@ -27,17 +29,32 @@ const (
...
+		return c.readBuf.Read(p)
+	}
+
 	_, message, err := c.Conn.ReadMessage()
 	if err != nil {
 		return 0, err
 	}
-	return copy(p, message), nil
 
+	// Fast path, no need to stage in the intermediate buffer
```
ipostelnik, line 48: You should copy as much as will fit into `p`, then write leftover (if any) into `c.readBuf`

websocket/websocket_test.go
```diff
@@ -1,18 +1,23 @@
...
 	"crypto/x509"
-	"github.com/rs/zerolog"
+	"fmt"
 	"io"
 	"math/rand"
 	"net/http"
 	"testing"
+	"time"
+
+	"github.com/rs/zerolog"
```
ipostelnik, line 14: Please keep 3rd party imports separate from project imports and sort. Your IDE or goimports can do this for you.

websocket/connection.go
```diff
@@ -49,6 +66,19 @@ func (c *GorillaConn) Write(p []byte) (int, error) {
 	return len(p), nil
 }
 
+// SetDeadline sets both read and write deadlines, as per net.Conn interface docs:
+// "It is equivalent to calling both SetReadDeadline and SetWriteDeadline."
+// Note there is no synchronization here, but the gorilla implementation isn't thread safe anyway
+func (c *GorillaConn) SetDeadline(t time.Time) error {
```
ipostelnik, line 72: Is this to make `GorillaConn` implement `net.Conn` interface? Was it not implemented properly before? Did it not matter until now?

benbuzbee: Yes it is; I am not sure being a full net.Conn was ever a goal or if you just wanted it to be an `io.ReadWriter` but it seems a shame not to add a the simple interface to make it a net.Conn

websocket/connection.go
benbuzbee: updated as requested

ipostelnik (approved):

Issue #327, opened 2021-03-03
`cloudflared service install` broken on macOS


adamfeldman:
Doc: https://github.com/cloudflare/cloudflare-docs/blob/production/products/cloudflare-one/src/content/connections/connect-apps/run-tunnel/run-as-service.md

### Steps to reproduce
1. `sudo cloudflared service install`
2. `tail -f /Library/Logs/com.cloudflare.cloudflared.err.log /Library/Logs/com.cloudflare.cloudflared.out.log`
3. Observe repeating line in error log ```Use `cloudflared tunnel run` to start tunnel MY_TUNNEL```

### Expected result: 
Seeing the usual tunnel startup info in the logs

### A fix that worked for me
Edit `/Library/LaunchDaemons/com.cloudflare.cloudflared.plist`

Original (runs `/usr/local/bin/cloudflared`)
```
                <array>
                        <string>/usr/local/bin/cloudflared</string>
                </array>
```
Fixed (runs `/usr/local/bin/cloudflared tunnel run`)
```
                <array>
                        <string>/usr/local/bin/cloudflared</string>
                        <string>tunnel</string>
                        <string>run</string>
                </array>
```

ref: https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPSystemStartup/Chapters/CreatingLaunchdJobs.html

adamfeldman:
This was originally filed here: https://github.com/cloudflare/cloudflare-docs/issues/791

adamfeldman:
I've attached all the files:
-  [config-sanitized.yml.zip](https://github.com/cloudflare/cloudflared/files/6124188/config-sanitized.yml.zip)
-  [BROKEN-com.cloudflare.cloudflared.plist.zip](https://github.com/cloudflare/cloudflared/files/6124198/BROKEN-com.cloudflare.cloudflared.plist.zip)
- [WORKING-com.cloudflare.cloudflared.plist.zip](https://github.com/cloudflare/cloudflared/files/6124200/WORKING-com.cloudflare.cloudflared.plist.zip)
- [logs.zip](https://github.com/cloudflare/cloudflared/files/6124206/logs.zip)
  - Attached logs are for the error condition, I did not attach logs for when it's working fine with my `.plist` edits



When the problem is occurring, I only get output in `/Library/Logs/com.cloudflare.cloudflared.err.log`, and no output in `/Library/Logs/com.cloudflare.cloudflared.out.log`.

[...]

PR #341, opened 2021-03-26
Fix macOS service installation launch template

allending:
The current version of cloudflared (2021.3.2 at time of commit)
requires additional arguments to start a tunnel. Running 'cloudflared'
on its own does not run a tunnel with the default configuration.

This is behavior that has changed from previous versions of cloudflared,
where no arguments were required for starting/running a tunnel with the
default configuration behavior.

This fixes https://github.com/cloudflare/cloudflared/issues/327.

### DIFF
--- a/cmd/cloudflared/macos_service.go
+++ b/cmd/cloudflared/macos_service.go
@@ -49,6 +49,8 @@ func newLaunchdTemplate(installPath, stdoutPath, stderrPath string) *ServiceTemp
 		<key>ProgramArguments</key>
 		<array>
 			<string>{{ .Path }}</string>
+			<string>tunnel</string>
+			<string>run</string>
 		</array>
 		<key>RunAtLoad</key>
 		<true/>


ytbigvinny27 (changes requested):
Delete all

ipostelnik (changes requested):
We have users that rely on running cloudflared in "client" mode which doesn't use these arguments. We need to continue to support this.

adamfeldman:
@ipostelnik could this modal behavior be put behind a config flag? Last I checked, when reporting this pain point in March, there was no documentation about running `cloudflared` as a service in "client" vs. "tunnel" mode. [Please see this comment on the related issue](https://github.com/cloudflare/cloudflared/issues/327#issuecomment-833485533). Thank you!

01SMD (approved):

joliveirinha:
closing due to age

adampatterson:
@joliveirinha  This is still an actual issue though. At least on MacOS

https://github.com/cloudflare/cloudflared/issues/327

Issue #364, opened 2021-04-30
vnc failing to connect


gnawhleinad:
_I think_ the error is a result of incorrect ingress rule parsing.

The server `cloudflared` is using `config.yaml`:

```
tunnel: redacted
credentials-file: /etc/cloudflared/redacted.json
origincert: /etc/cloudflared/cert.pem

ingress:
  - hostname: redacted.net
    path: /vnc/5901
    service: tcp://localhost:5901
... # multiple `/vnc/<port>` per user, there are 9 skipped here
  - hostname: redacted.net
    service: ssh://localhost:22
  - service: http_status:404
```

The server `cloudflared` is using the wrong ingress rule. It should be using ingress rule `0` but it is using `10`:

```
{"level":"debug","time":"2021-04-30T17:15:59Z","message":"CF-RAY: 64826a4e3bb0057d-LAX GET http://localhost:8080/favicon.ico HTTP/1.1"}
[...]

PR #365, opened 2021-04-30
debug: log host / path

gnawhleinad:
to help understand why the ingress rule logged is being selected.

to debug https://github.com/cloudflare/cloudflared/issues/364.

### DIFF
--- a/origin/proxy.go
+++ b/origin/proxy.go
@@ -238,8 +238,13 @@ func (p *proxy) logRequest(r *http.Request, fields logFields) {
 	} else {
 		p.log.Debug().Msgf("All requests should have a CF-RAY header. Please open a support ticket with Cloudflare. %s %s %s ", r.Method, r.URL, r.Proto)
 	}
-	p.log.Debug().Msgf("CF-RAY: %s Request Headers %+v", fields.cfRay, r.Header)
-	p.log.Debug().Msgf("CF-RAY: %s Serving with ingress rule %v", fields.cfRay, fields.rule)
+	p.log.Debug().
+		Str("CF-RAY", fields.cfRay).
+		Fields(r.Header).
+		Str("host", r.Host).
+		Str("path", r.URL.Path).
+		Interface("rule", fields.rule).
+		Msg("Inbound request")
 
 	if contentLen := r.ContentLength; contentLen == -1 {
 		p.log.Debug().Msgf("CF-RAY: %s Request Content length unknown", fields.cfRay)


origin/proxy.go
```diff
@@ -239,7 +239,7 @@ func (p *proxy) logRequest(r *http.Request, fields logFields) {
 		p.log.Debug().Msgf("All requests should have a CF-RAY header. Please open a support ticket with Cloudflare. %s %s %s ", r.Method, r.URL, r.Proto)
 	}
 	p.log.Debug().Msgf("CF-RAY: %s Request Headers %+v", fields.cfRay, r.Header)
-	p.log.Debug().Msgf("CF-RAY: %s Serving with ingress rule %v", fields.cfRay, fields.rule)
+	p.log.Debug().Msgf("CF-RAY: %s Serving (%s, %s) with ingress rule %v", r.Host, r.URL.Path, fields.cfRay, fields.rule)
```
sssilver, line 242: This is a good idea, but why not just add those fields as structured log values (and move the ingress rule %v into a structured log field as well)?

Something like:
```go
    p.log.Debug().
        Str("CF-RAY", fields.cfRay).
        Str("host", r.Host).
        Str("path", r.URL.Path).
        Str("rule", fields.rule).
        Msg("Inbound request")
```

This would be more idiomatic for what we would like to achieve.

sssilver (changes requested):
Thanks for the PR! Left a comment about structured logging.

gnawhleinad:
> This is a good idea, but why not just add those fields as structured log values

@sssilver, thanks for the feedback! I made the adjustment and opted to add the log as opposed to editing the existing one (in case some customers depended on its log format). I'm happy to remove the existing "Serving with ingress...", too.

sssilver:
> > This is a good idea, but why not just add those fields as structured log values
> 
> @sssilver, thanks for the feedback! I made the adjustment and opted to add the log as opposed to editing the existing one (in case some customers depended on its log format). I'm happy to remove the existing "Serving with ingress...", too.

Debug logs aren't a part of our public contract :) Let's roll both of these into the log statement you made:

```go
// Remove these
p.log.Debug().Msgf("CF-RAY: %s Request Headers %+v", fields.cfRay, r.Header)
p.log.Debug().Msgf("CF-RAY: %s Serving with ingress rule %v", fields.cfRay, fields.rule) 
```

Roll `r.Header` into the debug log context:

```go
	p.log.Debug().
		Str("CF-RAY", fields.cfRay).
		Str("Header", r.Header).  // not sure if Str is the right field type here
		Str("host", r.Host).
		Str("path", r.URL.Path).
		Str("rule", fields.rule).
		Msg("Inbound request")
```

sssilver (approved):

ipostelnik (approved):

PR #367, opened 2021-05-04
Setup a Github action for checking the cloudflared build

sssilver:


nmldiegues (approved):

chungthuang (approved):

.github/workflows/check.yaml
```diff
@@ -0,0 +1,20 @@
+on: [push, pull_request]
+name: Check
+jobs:
+  check:
+    strategy:
+      matrix:
+        go-version: [1.15.x]
```
sudarshan-reddy, line 7: Should we simply bump it to `1.16.x` while we are at it?

sudarshan-reddy (approved):

.github/workflows/check.yaml
sssilver: It's not a simple bump; compilation starts failing

PR #386, opened 2021-05-31
Allow TokenID and TokenSecret as env vars for cloudflares access

### LINKED ISSUE #232: Allow passing token id/secret using filesystem
We use Argo Tunnel and Cloudflare to access protected resources and as a handy way to let our end users access said resources we packaged `cloudflared` in a container so opening a tunnel is as easy as starting the docker service.

We use individual tokens to automate authentification and therefore pass the ID and Token as command parameters which is not the most ideal thing to do as you could access it by listing the host's processes : 
`/bin/cloudflared access tcp --id ${CF_ID} --secret ${CF_SECRET} ...`

It would be better to have the ability to read those bits of information from the file system like this : 
`/bin/cloudflared access tcp --id-file /secrets/my_id --secret-file /secrets/my_secret ...`

What do you think ?

Maartincm:
resolves #232

Feel free to ask for anything!

### DIFF
--- a/cmd/cloudflared/access/cmd.go
+++ b/cmd/cloudflared/access/cmd.go
@@ -143,13 +143,13 @@ func Commands() []*cli.Command {
 							Name:    sshTokenIDFlag,
 							Aliases: []string{"id"},
 							Usage:   "specify an Access service token ID you wish to use.",
-							EnvVars: []string{"CLOUDFLARE_SERVICE_TOKEN_ID"},
+							EnvVars: []string{"TUNNEL_SERVICE_TOKEN_ID"},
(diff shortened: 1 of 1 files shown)


chungthuang:
Thank you for contributing to the code base! May I ask that you change the env var prefix to `TUNNEL` to match the style in https://github.com/cloudflare/cloudflared/blob/master/cmd/cloudflared/tunnel/cmd.go#L677?

chungthuang (approved):

sudarshan-reddy (approved):

sssilver (approved):

PR #387, opened 2021-06-03
add CGO_ENABLED=0 in Makefile

wwqgtxx:
Disable cgo to independent of the system glibc could avoid `TLS handshake timeout` in old glibc platform (eg: debian8 on openvz7)  and work better in alpine based docker image.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -88,7 +88,7 @@ ifeq ($(FIPS), true)
 	mv fips/fips.go.linux-amd64 fips/fips.go
 endif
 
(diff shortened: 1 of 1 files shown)


sodabrew:
The Dockerfile has this set: https://github.com/cloudflare/cloudflared/blob/master/Dockerfile#L6

sudarshan-reddy:
I prefer this control to rest with the caller of Makefile. For example: https://github.com/cloudflare/cloudflared/blob/master/build-packages.sh#L3.

sodabrew:
Is there a reason you wrote build-packages.sh instead of `make packages` as a Makefile target with the same code?

sudarshan-reddy:
Probably opinionated. I felt it was easier to manage all package builds in one place and not add too many "scripty" things to the Makefile.

PR #470, opened 2021-09-18
Add Docker Workflow

### LINKED ISSUE #422: Docker tag latest isn't latest version
I am trying to integrate cloudflared's docker image into my CoreOS's instance, I want the newest version of cloudflared is always pulled, but when you do a pull of `latest` tag:
```
[core@fcos ~]$ podman run --rm docker.io/cloudflare/cloudflared:latest -v
cloudflared version 2020.7.0 (built 2020-07-01-1654 UTC)
```
Seems that you always have to sepcify the version like `docker.io/cloudflare/cloudflared:2021.7.0`. And always update the version string accordingly.
***
The root cause of this problem (I think) is that you did not tag latest version with `latest` tag, if doing so, pulling latest docker image will be convenient.

Anthony-Bible:
This adds a docker workflow that will build and push the images to GHCR and DockerHub. There are a few things to consider

1.  You will need to add `DOCKERHUB_USERNAME` and `DOCKERHUB_PASSWORD` consisting of your docker username and an authentication token to this repository's secrets
2.  This will only run when pushes are to master branch or a release is made (this includes pre-releases)
3. For docker tags, it uses either the tag (2021.9.0) or the git short sha. 
    a. This appears to be what's happening in the docker repo, so I tried replicating it. 
4. It will push to` ghcr.io/cloudflare/cloudflared` and `docker.io/cloudflare/cloudflared`
     

 This Fixes #422

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -6,21 +6,19 @@ on:
     branches:
       - master
 
-    # Publish `v1.2.3` tags as releases.
-    tags:
-      - v*
+  release:
+    types: [published]
 env:
-  BUILD_VERSION: "2021.9.0"
+  BUILD_VERSION: ${{ github.ref }}
   IMAGE_NAME: cloudflared
   DOCKER_CLI_EXPERIMENTAL: enabled
-  REPOSITORY: ${{ github.actor }}/${{ github.workflow }} 
+  REPOSITORY: ${{ github.repository }}
 
 jobs:
   # Push image to GitHub Packages.
   # See also https://docs.docker.com/docker-hub/builds/
   push:
-    # Ensure test job passes before pushing image.
-    runs-on: ubuntu:20.04
+    runs-on: ubuntu-latest
     if: ${{ github.event_name != 'pull_request' }}
     steps:
       - name: Source checkout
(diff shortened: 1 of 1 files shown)


nmldiegues:
Thanks for this contribution @Anthony-Bible 

I think we'd want this to only push images on tagged commits. That's what we have been doing thus far --- except we haven't tagged any with `latest`.
Is there a good reason to change that?

Anthony-Bible:
> I think we'd want this to only push images on tagged commits. That's what we have been doing thus far --- except we haven't tagged any with `latest`.
> Is there a good reason to change that?

Yep, looking at the docker images it looks like they're tagged with the commits the git tag is pointing too. I originally thought these were pushes to master also being built. I'll go ahead and change that.

nmldiegues:
@sssilver  would you be able to get the necessary secrets plugged into the repo so that they can be used from this workflow?

jauderho:
It appears to be customary to tag the latest release with the `latest` tag and probably helpful for folks that just want the most recent release. 

So unless there are specific objections to doing so, I would suggest leaving that in place.

nmldiegues:
> It appears to be customary to tag the latest release with the `latest` tag and probably helpful for folks that just want the most recent release.
> 
> So unless there are specific objections to doing so, I would suggest leaving that in place.

Isn't that being done here?
```
          tags: |
            ${{env.REPOSITORY}}:${{ env.app_version }}
            ${{env.REPOSITORY}}:latest
            ghcr.io/${{ env.REPOSITORY }}:${{ env.app_version }}
            ghcr.io/${{ env.REPOSITORY }}:latest
```

Anthony-Bible:
> > It appears to be customary to tag the latest release with the `latest` tag and probably helpful for folks that just want the most recent release.
> > So unless there are specific objections to doing so, I would suggest leaving that in place.
> 
> Isn't that being done here?
> 
> ```
>           tags: |
>             ${{env.REPOSITORY}}:${{ env.app_version }}
>             ${{env.REPOSITORY}}:latest
>             ghcr.io/${{ env.REPOSITORY }}:${{ env.app_version }}
>             ghcr.io/${{ env.REPOSITORY }}:latest
> ```

Yes it does push to latest and the tagged version of the docker image

.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,76 @@
...
+  # Push image to GitHub Packages.
+  # See also https://docs.docker.com/docker-hub/builds/
+  push:
+    runs-on: ubuntu-latest
+    if: ${{ github.event_name != 'pull_request' }}
+    steps:
+      - name: Source checkout
+        uses: actions/checkout@v2.3.4 
+ 
+      - name: Setup QEMU
```
sudarshan-reddy, line 22: Why do we need QEMU?

Anthony-Bible: Currently it isn't doing anything, but does lay the groundwork to build multi-architecture docker images if that is wanted. https://github.com/docker/build-push-action/blob/master/docs/advanced/multi-platform.md

Anthony-Bible: I do want to note, I didn't include the multi-architecture build when I submitted this PR because I noticed the docker image is only built for amd64. If this is something we want to incorporate I can always add it.

nmldiegues:
In the end it seems like we'll stick to the docker image being published from our internal systems. It should have the latest tag now.

PR #492, opened 2021-10-22
Add X-Forwarded-Host for http proxy

ibigbug:
Adding [`X-Forwarded-Host`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host) header to the http proxy.

This can be helpful in the case when the service in the ingress rule is another reverse proxy(call this 2nd reverse proxy), and the 2nd reverse proxy wants to pass original host to the backend application. So it becomes:

Original request:

```
mydns.on.cloudflare <---- customer request
        |
        |
cloudflared
  - hostname: mydns.on.cloudflare
    service: 2nd-proxy.some.internal
```

So in the `2nd-proxy.some.internal`, it listens ` 2nd-proxy.some.internal` and can read `$http_x_forwarded_host` and pass it to the application.

If the idea makes sense, maybe we want to add a new config option, as well as tests/docs update.

### DIFF
--- a/ingress/origin_proxy.go
+++ b/ingress/origin_proxy.go
@@ -40,6 +40,8 @@ func (o *httpService) RoundTrip(req *http.Request) (*http.Response, error) {
 
 	if o.hostHeader != "" {
 		// For incoming requests, the Host header is promoted to the Request.Host field and removed from the Header map.
+		// Pass the original Host header as X-Forwarded-Host.
+		req.Header.Set("X-Forwarded-Host", req.Host)
 		req.Host = o.hostHeader
 	}
 	return o.transport.RoundTrip(req)
--- a/ingress/origin_proxy_test.go
+++ b/ingress/origin_proxy_test.go
@@ -3,6 +3,7 @@ package ingress
 import (
 	"context"
 	"fmt"
+	"io/ioutil"
 	"net"
 	"net/http"
 	"net/http/httptest"
@@ -118,7 +119,9 @@ func TestHTTPServiceHostHeaderOverride(t *testing.T) {
 			w.WriteHeader(http.StatusSwitchingProtocols)
 			return
 		}
-		w.Write([]byte("ok"))
+		// return the X-Forwarded-Host header for assertions
+		// as the httptest Server URL isn't available here yet
+		w.Write([]byte(r.Header.Get("X-Forwarded-Host")))
 	}
 	origin := httptest.NewServer(http.HandlerFunc(handler))
 	defer origin.Close()
@@ -140,6 +143,10 @@ func TestHTTPServiceHostHeaderOverride(t *testing.T) {
 	resp, err := httpService.RoundTrip(req)
 	require.NoError(t, err)
 	require.Equal(t, http.StatusOK, resp.StatusCode)
+
+	respBody, err := ioutil.ReadAll(resp.Body)
+	require.NoError(t, err)
+	require.Equal(t, respBody, []byte(originURL.Host))
 
 }
 


ipostelnik:
This hostname already comes in the `Host` header. I don't see the need to duplicate it in another header.

ibigbug:
@ipostelnik sometimes we don't want the original `Host` header for example a intermediate reverse proxy receiving the request "1st-tier.service.domain" and pass the request to backend "2nd-tier.service.domain", and the backend want to know the original host `1st-tier.service.domain" by inspecting the x-forwarded-for header.

nmldiegues:
I think @ipostelnik 's question is:
- if we make this change in cloudflared, then we will be sending 2 headers with the same contents
- instead, it seems to make more sense to have your 2nd tier reverse proxy take the Host header, put it in X-Forwarded-For, and only then put itself in the Host header value

ipostelnik:
> the actual case is I'm use cloudflared as the "2nd tier reverse proxy", which should be responsible to put in the X-Forwarded-For to backend.

This does not belong in cloudflared. As I said above, we already pass the public hostname that the client requested in the `Host:` header. This PR just copies Host header into `X-Forwarded-Host` header, which is not adding any value.

I'm curious how you're using Cloudflare as 2nd tier reverse proxy without making requests to the hostname configured for your zone.

ibigbug:
@ipostelnik makes sense.

But consider the above config scenario:

```
ingress:
  - hostname: gitea.example.com  # public domain
    service: http://gitea.internal  # internal domain
```

there are two different hostnames here.

I haven't figured out how `cloudflared` process the `Host` header yet.

How ever as per the RFC, if you request "www.example.com" directly to the server, the `Host` header value should be `www.example.com`, say if we don't do anything.

In this particular scenario, I'm using `cloudflared` as a reverse proxy, where it listens and receives request on `gitea.example.com`, and at here, `cloudfalred` should see `Host` with value `gitea.example.com`.

it will then proxy pass the request to `gitea.internal`, and if we don't mutate the request, the `gitea.internal` should see `Host` with value `gitea.internal`.

but if we mutate the original request, and change it to `gitea.example.com`, this may cause issue, if the `gitea.internal` doesn't accept request with hostname `gitea.example.com` at all (like Nginx can tell what hostname it listens on, only for example).

So at `cloudflared` it shouldn't mutate the `Host` and should keep the original `gitea.interal` while passing the request.(not sure if this is what you were describing "we already pass the public hostname that the client requested")

so in the above case, we need to pass the original "public hostname" `gitea.example.com` to "gitea.internal", as `X-Forwareded-Host`, I think this is also the purpose this header was invented for.

but anyways, in the case, there are two hostnames: `gitea.internal` and `gitea.example.com` that we need to let the backend service know, if we only pass a single `Host`, it won't be enough anyways.

@ipostelnik what's your thoughts?

ipostelnik:
Your origin webserver should be configured for the *public* hostname, `gitea.example.com`  in your example. The internal hostname of your server is irrelevant here. This would be no different if you had exposed your server on the internet directly. The requests would have your public hostname in the host header.

ibigbug:
@ipostelnik I'm not getting what you are saying: 

> the internal hostname of your server is irrelevant here

the `cloudflared` need this to pass the request to right?

I'm able to access my internal service through `gitea.example.com` and the internal service need to see the hostname as `gitea.example.com` rather than the `gitea.internal` - could you please suggest me if `cloudflared` can do this?

thank you.

ibigbug:
I'd make myself clear:

the `gitea.internal` is relevant because it only accept request with hostname: `gitea.internal` at level 7 - if the request arrives the server IP but with a different `Host` header, it won't work. and this is quite common for one server hosting multiple services with different hostnames (similar to TLS SNI).

So to have `cloudflared` to be able to proxy the request, I have to tell `cloudflared` to use `gitea.internal` to send request to backend internal service.

In which case, the backend service need to know the request was originally made with `Host`: `gitea.example.com`, so this is where the `X-Forwarded-Host` header does the work.

please let me know if the scenario is still not clear enough.

ipostelnik:
cloudfalred will set `Host:` header of the request to your origin to `
gitea.example.com`. You can confirm this by using something like httpbin as
your origin and examining the requests. Run `docker run -p 8000:80
kennethreitz/httpbin` and then in another terminal run your tunnel via
`cloudflared tunnel run --url http://localhost:8000 NAME`. Assuming you've
routed the tunnel NAME to PUBLICHOSTNAME, you can examine request headers
by going to https://PUBLICHOSTNAME/headers

Hope this helps clarify how cloudflared works.

-Igor

ipostelnik:
You can override host header sent to the origin using a configuration
setting in the ingress rules. See
https://developers.cloudflare.com/cloudflare-one/connections/connect-apps/configuration/configuration-file/ingress#httphostheader.
Here's an example:

```
ingress:
  - hostname: gitea.example.com  # public domain
    service: http://gitea.internal  # internal domain    originRequest:
       httpHostHeader: gitea.internal
```


-Igor

ipostelnik:
Formatting was lost in email reply, here's the example configuration stanza:

```
ingress:
  - hostname: gitea.example.com  # public domain
    service: http://gitea.internal  # internal domain
    originRequest:
       httpHostHeader: gitea.internal
```

ibigbug:
thanks so much @ipostelnik and I see what you mean.

This is what I'm doing already with the `httpHostHeader`

and that still has a problem: setting the `httpHostHeader` make the request to `gitea.internal` with `Host`: `gitea.internal` - this will make sure the `gitea.internal` can send the response back, without which, the request won't be processed. but the original `Host`: `gitea.example.com` will be lost, right?

ibigbug:
checkout this httpbin: https://httpbin.xiaoba.me/headers

```
{
  "headers": {
    "Accept": "text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.9", 
    "Accept-Encoding": "gzip", 
    "Accept-Language": "en-US,en;q=0.9,zh-CN;q=0.8,zh;q=0.7", 
    "Cache-Control": "max-age=0", 
    "Cdn-Loop": "cloudflare", 
    "Cf-Connecting-Ip": "2405:6e00:1303:501:1ca5:20e6:1f88:3694", 
    "Cf-Ipcountry": "AU", 
    "Cf-Ray": "6ac8aca959b74f46-SYD", 
    "Cf-Visitor": "{\"scheme\":\"https\"}", 
    "Cf-Warp-Tag-Id": "30c8b335-7eb1-41c4-aabf-0a77fd7fae48", 
    "Connection": "keep-alive", 
    "Dnt": "1", 
    "Host": "gitea.internal", 
    "Sec-Ch-Ua": "\"Microsoft Edge\";v=\"95\", \"Chromium\";v=\"95\", \";Not A Brand\";v=\"99\"", 
    "Sec-Ch-Ua-Mobile": "?0", 
    "Sec-Ch-Ua-Platform": "\"Windows\"", 
    "Sec-Fetch-Dest": "document", 
    "Sec-Fetch-Mode": "navigate", 
    "Sec-Fetch-Site": "none", 
    "Sec-Fetch-User": "?1", 
    "Upgrade-Insecure-Requests": "1", 
    "User-Agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/95.0.4638.69 Safari/537.36 Edg/95.0.1020.44", 
    "X-Forwarded-Host": "httpbin.xiaoba.me"
  }
}
```

the value of `X-Forwarded-For` is the original `Host`, which needs to be passed to `gitea.internal`, and this is running the `cloudflared` with the patch in this PR

ibigbug:
without this patch, `cloudflared` can only pass one `Host` back to the backend service, it's either the original `Host` or the `httpHostHeader`.

but actually in the real scenario, it needs both to be passed.

ipostelnik:
Ok, this makes more sense now. You want to preserve original Host header when applying host header override. I think that would  be ok. This should happen in https://github.com/cloudflare/cloudflared/blob/master/ingress/origin_proxy.go#L43

ibigbug:
thanks @ipostelnik updated as requested and tests as well

please take a look

nmldiegues (approved):

sudarshan-reddy (approved):

joliveirinha (approved):

chungthuang (approved):

ipostelnik (commented):
LGTM. Please squash into a single commit.

ibigbug:
@ipostelnik done

sssilver (approved):

ibigbug:
@ipostelnik are we good to merge?

ibigbug:
hello - what do I need to do to get this checked in ?

happy holidays :)

ipostelnik (approved):

ibigbug:
thanks - Merging is blocked
The base branch restricts merging to authorized users.

who can help to merge this?

PR #591, opened 2022-03-03
Fix log message

robinbraemer:
printing `seconds` is superfluous since time.Duration already adds the `s` suffix

Invalid log message would be
```
Retrying connection in up to 1s seconds
```


nmldiegues (approved):
Thanks for the fix 👍

DevinCarr (approved):

ipostelnik (approved):

sudarshan-reddy:
@robinbraemer would you mind rebasing this? 🙂

joliveirinha (approved):

PR #606, opened 2022-03-22
Addressing h2mux and session flaky tests

DevinCarr:
Updating some tests to try and appease CI flaky tests.

- Correct some h2mux tests on windows without CGO_ENABLED: Compression isn't supported without cgo on windows environments so the h2mux tests need to ignore compression.
   - TestMultipleStreamsWithDictionaries, TestSampleSiteWithDictionaries, TestLongSiteWithDictionaries
- TestReadFromDstSessionPreventClosed occasionally will fail because the session closes in the idle time. This might be because of all the tests running in parallel in separate goroutines and the 100ms time limit is exceeded. Bumping this idle time and execution window up may help reduce the flaky failures for these tests

### DIFF
--- a/h2mux/h2mux_test.go
+++ b/h2mux/h2mux_test.go
@@ -823,7 +823,7 @@ func TestMultipleStreamsWithDictionaries(t *testing.T) {
 		}
 
 		originMuxMetrics := muxPair.OriginMux.Metrics()
-		if q > CompressionNone && originMuxMetrics.CompBytesBefore.Value() <= 10*originMuxMetrics.CompBytesAfter.Value() {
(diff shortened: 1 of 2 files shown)


h2mux/h2mux_test.go
```diff
@@ -823,7 +823,7 @@ func TestMultipleStreamsWithDictionaries(t *testing.T) {
 		}
 
 		originMuxMetrics := muxPair.OriginMux.Metrics()
-		if q > CompressionNone && originMuxMetrics.CompBytesBefore.Value() <= 10*originMuxMetrics.CompBytesAfter.Value() {
+		if q > CompressionNone && CompressionIsSupported() && originMuxMetrics.CompBytesBefore.Value() <= 10*originMuxMetrics.CompBytesAfter.Value() {
```
chungthuang, line 826: Nit: The compression check can be grouped into its own function.

chungthuang (approved):

PR #610, opened 2022-04-03
Output QR Code to Make Login Easier

sl33nyc:
I'd like to do a `cloudflared tunnel login` in a situation where I don't have a browser handy.  Adding a QR code URL in the terminal output makes it simpler (for me and maybe others) to authorize the tunnel login on my phone (for example) without having to manually type in the URL on a different machine.

* I somewhat arbitrarily chose `github.com/mdp/qrterminal/v3` as the QR Code generator and am happy to switch it to something more amenable license, style, etc.-wise.
* Since no existing tests failed, I assume that there aren't any existing tests that verify the terminal output, which now includes the QR code.  :man_shrugging: 

cloudflared_qrcode

### DIFF
--- a/cmd/cloudflared/tunnel/login.go
+++ b/cmd/cloudflared/tunnel/login.go
@@ -23,13 +23,22 @@ const (
 	callbackStoreURL = "https://login.cloudflareaccess.org/"
 )
 
+var (
+	outputQrcodeFlag = &cli.BoolFlag{
+		Name:    "output-qr-code",
+		Usage:   "Output a QR Code containing a link to tunnel login authorization workflow",
+		Aliases: []string{"qr"},
+	}
+)
+
 func buildLoginSubcommand(hidden bool) *cli.Command {
 	return &cli.Command{
 		Name:      "login",
 		Action:    cliutil.ConfiguredAction(login),
 		Usage:     "Generate a configuration file with your login details",
-		ArgsUsage: " ",
+		UsageText: "cloudflared tunnel [tunnel command options] login [subcommand options]",
 		Hidden:    hidden,
+		Flags:     []cli.Flag{outputQrcodeFlag},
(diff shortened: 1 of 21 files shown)


nmldiegues:
> ```
> === RUN   TestReadFromDstSessionPreventClosed
>     session_test.go:181: 
>         	Error Trace:	session_test.go:181
>         	            				session_test.go:110
>         	Error:      	Received unexpected error:
>         	            	session closed while it's still active
>         	Test:       	TestReadFromDstSessionPreventClosed
> ```
> 
> I didn't look too closely, but flaky test?

Likely, yeah.

We have internal CI that runs the same tests happily. Haven't dedicated time to figuring out why there's flakiness here specifically.

sudarshan-reddy:
Thanks for this idea. Seems cool. I'd suggest making this an optional feature controlled by passing a flag.

nmldiegues:
There may be some flakiness in the Windows unit tests around some legacy part of the code (h2mux). That's what failed already twice;  I just rebuilt it, but that should not block merges.

sudarshan-reddy (approved):

DevinCarr (approved):

sssilver:
Hi! I have an alternative suggestion.

In the spirit of the UNIX philosophy of building tools that "do one thing, and do it well", why not just use `qrencode` and `grep`?

1. Make sure `grep` and `qrencode` are installed
2. `cloudflared login 2>&1 | grep -m1 "https://" | qrencode -o - -t ansiutf8`
3. PROFIT!!!!

image

One could adjust the `grep` pattern to have be more specific, but the idea remains the same.

I'll close the PR for now, please feel free to open an issue with a feature suggestion, if you would like to continue discussing this.

sl33nyc:
:cry:

PR #627, opened 2022-04-18
Set ForceAttemptHTTP2 on http.Transport to auto support HTTP/2

### LINKED ISSUE #625: Tunnel: support HTTP/2 to origin servers
**Describe the bug**
Cloudflared disabled support for HTTP/2 connection to origin servers; cause it sets the `TLSClientConfig` to a non-nil value on `http.Transport` when instantiating one; particularly in here :point_down: https://github.com/cloudflare/cloudflared/blob/8a07a900fde36efe48fd27ad5f974273015e8270/ingress/origin_service.go#L279-L287

> See Go `http` documentation on `TLSClientConfig`: https://github.com/golang/go/blob/go1.17.9/src/net/http/transport.go#L165-L169:
> ```go
>     // TLSClientConfig specifies the TLS configuration to use with
>     // tls.Client.
>     // If nil, the default configuration is used.
>     // If non-nil, HTTP/2 support may not be enabled by default.
>     TLSClientConfig *tls.Config
> ```

This can be easily fixed by setting [`ForceAttemptHTTP2`](https://github.com/golang/go/blob/go1.17.9/src/net/http/transport.go#L278-L283) to true on on the `http.Transport`; pretty assured that it wouldn't break anything, right?
> [`http.DefaultTransport`](https://github.com/golang/go/blob/go1.17.9/src/net/http/transport.go#L43-L54) also does this, which is used in `http.DefaultClient`.

**To Reproduce**
1. Create and run a simple HTTP server that supports HTTP/1.1 and HTTP/2 protocols (possibly with a mock TLS certificate); log the used HTTP protocol on each request.
2. Run cloudflared and create a tunnel to the HTTP server (possibly with `--no-tls-verify` option).
3. Issue a request to the server.
4. See `HTTP/1.1` log out (cloudflared never attempts HTTP/2 nor protocol upgrade from HTTP/1.1).

**Expected behavior**
HTTP/2 should be preferred to HTTP/1.1 when available (in this context, when connecting to origin servers).

**Environment and versions**
 - OS: Ubuntu 20.04.4 LTS
 - Architecture: x86_64
 - Version: 8a07a900fde36efe48fd27ad5f974273015e8270 (latest commit on master at the time).

**Additional context**
- I did the suggested fix, and it worked as expected (preferred HTTP/2 when available, otherwise sticked to HTTP/1.1); can make a PR if my judgement is relevant.
- Supporting HTTP/2 clear-text (`h2c`) connection to origin servers is a bit more complicated, but can be a feature request.
- I assumed that this is not the intended behavior, therefore issued a bug, not a feature request.

agcom:
[`ForceAttempHTTP2` doc](https://github.com/golang/go/blob/go1.17.9/src/net/http/transport.go#L278-L283):
```go
    ...
    // ForceAttemptHTTP2 controls whether HTTP/2 is enabled when a non-zero
    // Dial, DialTLS, or DialContext func or TLSClientConfig is provided.
    // By default, use of any those fields conservatively disables HTTP/2.
    // To use a custom dialer or TLS config and still attempt HTTP/2
    // upgrades, set this to true.
    ForceAttemptHTTP2 bool
    ...
```

Found two instances of `http.Transport` instantiations where it needed to set `ForceAttemptHTTP2` boolean, in order to automatically enable HTTP/2 support; others are not needy, according to docs.

> Just to mention, there is also the [`http2.ConfigureTransport`](https://pkg.go.dev/golang.org/x/net/http2#ConfigureTransport) function that enables HTTP/2 support for an `http.Transport` instance, the **manual** way; it's used once in the [Cloudflare API base client](https://github.com/cloudflare/cloudflared/blob/8a07a900fde36efe48fd27ad5f974273015e8270/cfapi/base_client.go#L71).

Closes #625.

### DIFF
--- a/ingress/origin_service.go
+++ b/ingress/origin_service.go
@@ -284,6 +284,7 @@ func newHTTPTransport(service OriginService, cfg OriginRequestConfig, log *zerol
 		TLSHandshakeTimeout:   cfg.TLSTimeout.Duration,
 		ExpectContinueTimeout: 1 * time.Second,
 		TLSClientConfig:       &tls.Config{RootCAs: originCertPool, InsecureSkipVerify: cfg.NoTLSVerify},
+		ForceAttemptHTTP2:     true,
 	}
 	if _, isHelloWorld := service.(*helloWorld); !isHelloWorld && cfg.OriginServerName != "" {
 		httpTransport.TLSClientConfig.ServerName = cfg.OriginServerName
--- a/tunneldns/https_upstream.go
+++ b/tunneldns/https_upstream.go
@@ -135,6 +135,7 @@ func configureClient(hostname string, maxUpstreamConnections int) *http.Client {
 		MaxIdleConns:       1,
 		MaxConnsPerHost:    maxUpstreamConnections,
 		Proxy:              http.ProxyFromEnvironment,
+		ForceAttemptHTTP2:  true,
 	}
 	_ = http2.ConfigureTransport(transport)
 


nmldiegues:
So I dug a little further and I found out that we already tried to do this in the past: TUN-3896

We rolled it back because we had reports where cloudflared was error-ing when reaching out to HTTP origins, with errors such as:
```
Error proxying request to origin: stream error: stream ID 3; INTERNAL_ERROR
```

which seemed to be that the http2 golang client was not able to correctly negotiate the protocol to use with the origin.


We reverted it in TUN-4153 and never looked back to it.

Not sure how we can proceed, but this is why I had that hunch to ping you @sudarshan-reddy

agcom:
IMHO HTTP/2 benefits out-weights a rare error. A possible safe solution would be to disable HTTP/2 intentionally, and only enable it through a config (e.g. `--http2-origin`).

nmldiegues:
> IMHO HTTP/2 benefits out-weights a rare error. A possible safe solution would be to intentionally disable HTTP/2, and only enable it through a config (e.g. `--http2-origin`).

SGTM

That'd be under `OriginRequestConfig` (which shows up in more than 1 place because this is remotely configurable via https://blog.cloudflare.com/ridiculously-easy-to-use-tunnels/ ) if you want to take a stab at it

ssesik:
Hi, I tested it by patching cloudflared and it works for GET requests but not POST.

`error="Unable to reach the origin service. The service may be down or it may not be responding to traffic from cloudflared: http2: Transport: cannot retry err [stream error: stream ID 111; PROTOCOL_ERROR; received from peer] after Request.Body was written; define Request.GetBody to avoid this error" cfRay="id" ingressRule=0 originService=unix+tls:///path/to/service.sock`
Web server is nginx 1.20.1 php-fpm 8.1

Would really appreciate any further development on this.

PR #656, opened 2022-06-01
Add Http2Origin option to force HTTP/2 origin connections

nikr-canva:
From `cloudflared` issue #625 

Introduces a `http2Origin` option (and `--http2-origin` command line arg) to set the `ForceAttemptHTTP2` HTTP Transport option for origin services.

### DIFF
--- a/ingress/config.go
+++ b/ingress/config.go
@@ -227,6 +227,9 @@ func originRequestFromConfig(c config.OriginRequestConfig) OriginRequestConfig {
 			}
 		}
 	}
+	if c.Http2Origin != nil {
+		out.Http2Origin = *c.Http2Origin
+	}
 	return out
 }
 
@@ -377,6 +380,12 @@ func (defaults *OriginRequestConfig) setIPRules(overrides config.OriginRequestCo
 	}
 }
 
+func (defaults *OriginRequestConfig) setHttp2Origin(overrides config.OriginRequestConfig) {
+	if val := overrides.Http2Origin; val != nil {
+		defaults.Http2Origin = *val
+	}
+}
+
 // SetConfig gets config for the requests that cloudflared sends to origins.
 // Each field has a setter method which sets a value for the field by trying to find:
 //   1. The user config for this rule
@@ -402,6 +411,7 @@ func setConfig(defaults OriginRequestConfig, overrides config.OriginRequestConfi
 	cfg.setProxyAddress(overrides)
 	cfg.setProxyType(overrides)
 	cfg.setIPRules(overrides)
+	cfg.setHttp2Origin(overrides)
 	return cfg
 }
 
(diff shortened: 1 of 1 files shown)


nmldiegues:
The change LGTM

I suppose that if/when we want to force HTTP2 all the time, and just let people opt-out, we can add a new flag (to opt-out) and this new one will be moot. If everyone agrees, then I'm ok to go with this.

nmldiegues (approved):

nmldiegues:
There's just 1 little note here (for the team here): we'll have to get this new flag on the Ingress config in the Tunnel UI, right?

nikr-canva:
Yep, I missed a test. I'll get onto fixing it today.

chungthuang:
> 

Exactly.

chungthuang:
> There's just 1 little note here (for the team here): we'll have to get this new flag on the Ingress config in the Tunnel UI, right?

Exactly.

chungthuang (approved):

nikr-canva:
One additional comment on this - The original PR (#627) that I based this off also set the `ForceAttemptHTTP2` option in `tunneldns/https_upstream.go:configureClient()`, but I believe that is unrelated to the origin service connection (and rather is for DoH) so I haven't included that change in here. 

Let me know if I've misread the purpose of that package.

nmldiegues:
> but I believe that is unrelated to the origin service connection (and rather is for DoH) so I haven't included that change in here.

Correct.
I do not have opinions on the DNS proxy part. I'd rather keep it as is if no one is interested in evolving it.

nmldiegues:
There's just 1 tiny thing bit left:
- in ingress/config.go#setConfig , you need to set an override for this new property (this is used for getting the config from Cloudflare API and overriding any local properties after cloudflared starts, in case it is managed "remotely" as per https://blog.cloudflare.com/ridiculously-easy-to-use-tunnels/ )
- in the same file, then do the opposite in ConvertToRawOriginConfig

nikr-canva:
> in ingress/config.go#setConfig , you need to set an override for this new property (this is used for getting the config from Cloudflare API and overriding any local properties after cloudflared starts, in case it is managed "remotely" as per https://blog.cloudflare.com/ridiculously-easy-to-use-tunnels/ )
    in the same file, then do the opposite in ConvertToRawOriginConfig

Yeah I just spotted that as well, when I was looking through the test failures (or flakes?).

I'll update the override stuff.

nmldiegues:
> (or flakes?)

I don't know why, but the unit tests flake here (in GH CI); 
We run our own internal CI (also across Win/Linux/Mac) and they run just fine. So we never spent much time investigating it.

So no, there's no test failing because of the things lacking in https://github.com/cloudflare/cloudflared/pull/656#issuecomment-1144543977
There should be, but I'm not going to ask you to do that :)

nmldiegues:
> I'll update the override stuff.

Just to make sure, we're just pending on that for this to be merged.

nikr-canva:
> Just to make sure, we're just pending on that for this to be merged.
@nmldiegues ACK, I will aim to finish this up in the next day or so.

nmldiegues:
Cool, thanks.

We'll get it merged.

PR #618, opened 2022-04-08
feat(docker): build docker images with buildkit, add actions

initdc:
hello, I add docker build workflow for cloudflared. I am learning buildx near time, and I feel good.

It may also helpful in your bin distro.

check more usage from: https://github.com/initdc/ssh-proxy-by-caddy-l4

PR #681, opened 2022-06-30
Fix: Replace TARGET_GOOS and TARGET_GOARCH with TARGET_OS and TARGET_ARCH

NicolasDorier:
There is a mismatch between the environment name in the docker file and what the makefile is expecting:
* https://github.com/cloudflare/cloudflared/blob/master/Makefile#L78
* https://github.com/cloudflare/cloudflared/blob/master/Makefile#L69

You also need to declare the `ARG` inside the `FROM` if you intend to use it in the build stage. See https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,10 +2,12 @@
 ARG TARGET_GOOS
 ARG TARGET_GOARCH
 FROM golang:1.17.1 as builder
+ARG TARGET_GOOS
+ARG TARGET_GOARCH
 ENV GO111MODULE=on \
     CGO_ENABLED=0 \
-    TARGET_GOOS=${TARGET_GOOS} \
-    TARGET_GOARCH=${TARGET_GOARCH}
+    TARGET_OS=${TARGET_GOOS} \
+    TARGET_ARCH=${TARGET_GOARCH}
     
 LABEL org.opencontainers.image.source="https://github.com/cloudflare/cloudflared"
 


SayakMukhopadhyay:
I think this PR might tie into #618 which reworks the creation of the containers to make it multi platform. That PR and the related issue has been sitting for a while now though.

sudarshan-reddy:
The ARGS are already declared in lines 2 and 3.

And the Environment variables are correct. Notice `make cloudflared`. It uses these values: https://github.com/cloudflare/cloudflared/blob/master/Makefile#L78

sudarshan-reddy (changes requested):
The ARGS are already defined in lines 2 and 3. This PR adds them redundantly. 

And The environment variables are used here: https://github.com/cloudflare/cloudflared/blob/master/Makefile#L78

sudarshan-reddy:
@NicolasDorier I'm going to close this PR because of the reason I stated above. However, do not let this discourage you from opening more PRs. This codebase can be so much better and we will take all the amazing help we can get! ❤️

sudarshan-reddy:
Hmm thanks so much for leaving this comment. I've made a mistake. I thought this PR was adding a non-existent variable when in fact it was removing it.

Dockerfile
```diff
@@ -2,10 +2,12 @@
...
 ARG TARGET_GOARCH
 FROM golang:1.17.1 as builder
+ARG TARGET_GOOS
+ARG TARGET_GOARCH
 ENV GO111MODULE=on \
     CGO_ENABLED=0 \
-    TARGET_GOOS=${TARGET_GOOS} \
-    TARGET_GOARCH=${TARGET_GOARCH}
+    TARGET_OS=${TARGET_GOOS} \
+    TARGET_ARCH=${TARGET_GOARCH}
```
SayakMukhopadhyay, line 10:
```suggestion
    GOOS=${TARGET_GOOS} \
    GOARCH=${TARGET_GOARCH}
```
I think that instead of setting the `TARGET_OS` and `TARGET_ARCH` env vars in here, we should set the `GOOS` and `GOARCH` env vars as that is what is getting used primarily in the Makefile which in turn are getting used to set `TARGET_OS` and `TARGET_ARCH`.
https://github.com/cloudflare/cloudflared/blob/1733fe8c656c6104547684367f3a70899e790402/Makefile#L53-L57
https://github.com/cloudflare/cloudflared/blob/1733fe8c656c6104547684367f3a70899e790402/Makefile#L76-L80
So, directly setting `TARGET_OS` and `TARGET_ARCH` might have some side effects.

SayakMukhopadhyay (changes requested):

initdc:
We may met the same problem.
https://github.com/AdguardTeam/dnsproxy/pull/228/files#diff-ea8ccbc8293bff3932d8cc333b23b76ce61269aef2b38e253e5335854f340b40R14

SayakMukhopadhyay:
> We need make one thing clear: `docker buildx` does not have the ability to create cross-platform binaries because it only takes Linux as an argument.
> 
> > https://github.com/initdc/ssh-proxy-by-caddy-l4/blob/build/sh/README.md?plain=1#L50
> 
> We need to build an array of target platforms to loop runing `go build`.
> 
> > https://github.com/initdc/distro-arch-rootfs/blob/master/gen-chart.rb#L10

That's correct! That's what your PR #618 is doing. I believe this PR is loosely relates to that and does nothing to fix the problem of multi build, only fixes the env vars as a temporary fix, at least that's what I think it is.

NicolasDorier:
> Can we remove the args defined in line 2 and 3?

No you can't

NicolasDorier:
> Can we remove the args defined in line 2 and 3?

I believe that at tried and it didn't work.

Just to let you know: We are building our own cross platform cloudflared image on [here](https://github.com/btcpayserver/dockerfile-deps/tree/master/Cloudflared/2022.6.3).
The docker image is **btcpayserver/cloudflared:2022.6.3**, and we already use it on several architectures based on this fix.
I sadly don't have the time to test variations, so feel free to close this PR and supersede it with another one.

I believe it would be better to fix that in a deeper way: Rather than using `TARGET_GOOS`, just using the official `GOOS`.

sudarshan-reddy:
Im closing this as we now have docker images that this PR tried to achieve.

Issue #563, opened 2022-02-06
Proxy to a different path on the origin service


brenner-tobias:
**Is your feature request related to a problem? Please describe.**
In the ingress rules, it would be great to have the option to define a location of the service and not "only" a URL limited to protocol, hostname and IP (for http), e.g. `http://192.168.178.1/nas`

**Describe the solution you'd like**
I do not think a new option is needed, it could just be a part of the `service` URL.

**Describe alternatives you've considered**
Could also be added using an additional variable, like it is done for `path` on the hostname.

**Additional context**
If I add a URL with a location attached, I get the error:
`http://192.168.178.1/nas is an invalid address, ingress rules don't support proxying to a different path on the origin service. The path will be the same as the eyeball request's path`

nmldiegues:
Hello @brenner-tobias ,

Have you checked https://developers.cloudflare.com/cloudflare-one/connections/connect-apps/configuration/configuration-file/ingress ?
I believe the `path` configuration of a service does what you are mentioning.

nmldiegues:
That's much clearer, thanks!

That's not possible, indeed.
Of course you can set that up by running a nginx or equivalent in localhost and doing those transformations.
cloudflared is not aiming to be a general proxy, but we'll keep this in mind should it overlap with something we can provide of value

regnaio:
This proposed feature would be important and can help eliminate several DNS records for the same tunnel, e.g.

```yaml
ingress:
    - hostname: subdomain1.domain.com
      service: https://localhost:3001
    - hostname: subdomain2.domain.com
      service: https://localhost:3002
    - hostname: subdomain3.domain.com
      service: https://localhost:3003
    # etc...
```

regnaio:
Can a `cloudflared` contributor please advise?

To implement this feature, should we use [ProxyHTTP function](https://github.com/cloudflare/cloudflared/blob/master/proxy/proxy.go#L61) as a starting point?

nmldiegues:
> Can a `cloudflared` contributor please advise?
> 
> To implement this feature, should we use [ProxyHTTP function](https://github.com/cloudflare/cloudflared/blob/master/proxy/proxy.go#L61) as a starting point?

Yes, that is the starting point.

regnaio:
@brenner-tobias @nmldiegues Thanks for the ideas!

I came up with a rough working proof-of-concept

https://github.com/cloudflare/cloudflared/compare/master...regnaio:regnaio/location

```yaml
ingress:
[...]

PR #706, opened 2022-07-25
Regnaio/location

chippytech:
as


sudarshan-reddy:
Hello @chippytech ! what is the intention behind this PR?

chippytech:
To add location to cloudflared.

sudarshan-reddy:
I would like to suggest you be more descriptive.

What does this change do?

What warranted it?

Have you tested it? Do you have  demonstrations?

Why do you think this will benefit others?

Once we have details like this, we can debate if we want to add this to `cloudflared`

chippytech:
To add location to cloudflared

sudarshan-reddy:
I'm sorry @chippytech . Without the information asked for it is impossible to understand and deliberate about the merits of this PR. I suggest reopening this when you have this information. 🙂

chippytech:
To add location to cloudflared config.

fabiopires:
This PR is not very useful, but the functionality that it's trying to merge is.

The original author has described [here](https://github.com/cloudflare/cloudflared/issues/563#issuecomment-1081779878) what is the goal behind this merge. In fact, it will address the possibility of manipulating the location to which a request should go.

Hopefully this can be re-opened and potentially merged.

Issue #119, opened 2019-07-16
Support for XDG Base Directory spec. (aka stop filling my ~ up)


frebib:
https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html

We should be trying for `$XDG_CONFIG_HOME/cloudflared` first instead of `~/.cloudflared` etc. This is a pretty well accepted standard by now. Please consider revising how the directories are structured

lhanson:
https://github.com/OpenPeeDeeP/xdg makes it dead simple to respect users' preferences, please don't create arbitrary directories in my `$HOME`.

SmirnovAlexander:
+1
is it solved?

mintyfrankie:
+1
Would really love to see this happen -- `.cloudflared` is for now the only directory that I have to deal with before having a clean home directory.

TheDSCPL:
+1
Please implement this. XDG compliance is such a nice thing and most software is now compliant. It's a shame cloudflared still isn't

mbelyayev:
+1. XDG compliance is a standard these days.

PR #731, opened 2022-08-17
configuration: Follow XDG Base Directory standard for appropriate environments

Leptopoda:
imlements: #119

### What it does:
Follows the XDG Base Directory standard for the user config dir.
It uses [github.com/adrg/xdg](https://github.com/adrg/xdg) to find the right dir. It will try to resolve the `$XDG_CONFIG_HOME` and revert to the platform specific default ones.

Prior directories will still work (similar to the old windows `~/cloudflare-warp`).
I've implemented it on all Platforms but don't know how used they are on OSX or Windows so if desired I'll only enable it on Linux.

### What it doesn't:
It does not implement the xdg spec regarding `/etc/xdg/cloudflared` as It isn't commonly used, with the only big software using it on my system being systemd which symlinks back to /etc/systemd. Further the dir isn't used at all on any of my servers which I guess is the primary client of `/etc/cloudfalre`.
Also the issue did mention the cleanup of `$HOME`and not the full implementation of the spec ;)

#### Links
- https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

### DIFF
--- a/config/configuration.go
+++ b/config/configuration.go
@@ -30,6 +30,17 @@ var (
 	// DefaultUnixLogLocation is the primary location to find log files
 	DefaultUnixLogLocation = "/var/log/cloudflared"
 
+	// Default paths and env to determin the correct user config dir
+	LinuxUserConfigDirEnv   = "XDG_CONFIG_HOME"
+	LinuxUserConfigDir      = "~/.config"
+	WindowsUserConfigDirEnv = "%LOCALAPPDATA%"
+	WindowsUserConfigDir    = "~/AppData/Local"
+	DarwinUserConfigDir     = "~/Library/Application Support"
+
+	// App and Vendor ID used in the filepath
+	AppID    = "cloudflared"
+	VendorID = "cloudflare"
+
 	// Launchd doesn't set root env variables, so there is default
 	// Windows default config dir was ~/cloudflare-warp in documentation; let's keep it compatible
 	defaultUserConfigDirs = []string{"~/.cloudflared", "~/.cloudflare-warp", "~/cloudflare-warp"}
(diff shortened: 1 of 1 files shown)


config/configuration.go
```diff
@@ -30,9 +31,12 @@ var (
 	// DefaultUnixLogLocation is the primary location to find log files
 	DefaultUnixLogLocation = "/var/log/cloudflared"
 
+	// DefaultXDGConfigLocation is the primary user location to find a config file
+	DefaultXDGConfigLocation = xdg.ConfigHome + "/cloudflared"
```
sudarshan-reddy, line 35: Can we do this without adding a vendor to just bring in this variable?

config/configuration.go
```diff
@@ -30,9 +31,12 @@ var (
...
+
 	// Launchd doesn't set root env variables, so there is default
 	// Windows default config dir was ~/cloudflare-warp in documentation; let's keep it compatible
-	defaultUserConfigDirs = []string{"~/.cloudflared", "~/.cloudflare-warp", "~/cloudflare-warp"}
+	defaultUserConfigDirs = []string{DefaultXDGConfigLocation, "~/.cloudflared", "~/.cloudflare-warp", "~/cloudflare-warp"}
```
sudarshan-reddy, line 39: This should only be appended to dirs if the environment is xdg

sudarshan-reddy (changes requested):

config/configuration.go
Leptopoda: the xdg spec says that if the enviornment variable is not set the default location of `$HOME/.config` should be used.
Many linux distros/users do not set the env and expect this behavior.

Thats also the primary reason I used the lib and do not read directly from the env as every platform specifies its own standard dir. Checkout the libs readme to see the ones used.

config/configuration.go
Leptopoda: what do you men by that?

We need to specify a folder name for within `${XDG_CONFIG_HOME:-$HOME/.config}`
It will result in something similar to:
```
~/.config/cloudflared
├── XXX-XXX-XXX-XXX-XXX.json
└── cert.pem
```

sudarshan-reddy: The two comments are connected. You are relying on this library to populate the xdgconfig location prefix and my suggestion is to do that in the PR here and not bloat the the package with a library for just this logic.

Leptopoda: Just asking should the behavior also be implemented on OSX and Windows?
The appropriate dirs are `~/Library/Application Support` and `LocalAppData` but I don't know how popular they are?

Leptopoda:
Friendly ping :upside_down_face:

sudarshan-reddy (approved):

01SMD (approved):

chungthuang:
@Leptopoda thank you for your contribution. However, we are planning to move more configurations to be managed remotely, via the dashboard or API.

PR #759, opened 2022-09-19
Issue #672: Run cloudflared service without root

### LINKED ISSUE #672: Cloudflared service runs as root 🐛
**Describe the bug**
The `cloudflared service install` creates a service file that runs the `cloudflared` process as root. Unless some privileged operation is required (e.g. raw sockets), the service should run as a non-privileged user, such as `cloudflared`. Even with trusted processes, bugs that allow privilege escalation can appear.

**To Reproduce**
1. Run `cloudflared service install`;
2. A service file that starts the `cloudflared` process as root is created;
3. The cloudflare tunnel token is also visible in the service file (also reported by #666).

**Expected behavior**
The cloudflare service runs as a non-privileged user, such as `cloudflared`. Additional restrictions, such as SELinux policies (e.g., similar to `nginx`), should also improve security.

**Environment and versions**
 - OS: Centos Stream 8
 - Architecture: AMD64 (should not be architecture-dependent)
 - Version: 2022.6.2

**Additional context**
The issue may be resolved by adding `User=cloudflared` to the `Service` section of the configuration file. The user would have to be created, likely as a system user.

romeroyonatan:
Run the cloudflared service with `cloudflared` user

* Create user `cloudflared` if it does not exist in `cloudflared service install`
* Remove user `cloudflared` in `cloudflared service uninstall`
* Add `User=cloudflared` in `/etc/systemd/system/cloudflared.service`

Note: this only works for systemd.

I tested it in our environment and it worked :)

cloudflared-service-up

Closes #672

### DIFF
--- a/cmd/cloudflared/linux_service.go
+++ b/cmd/cloudflared/linux_service.go
@@ -44,6 +44,7 @@ const (
 	serviceCredentialFile = "cert.pem"
 	serviceConfigPath     = serviceConfigDir + "/" + serviceConfigFile
 	cloudflaredService    = "cloudflared.service"
+	cloudflaredUser       = "cloudflared"
 )
 
 var systemdTemplates = []ServiceTemplate{
@@ -59,6 +60,7 @@ Type=notify
 ExecStart={{ .Path }} --no-autoupdate{{ range .ExtraArgs }} {{ . }}{{ end }}
 Restart=on-failure
 RestartSec=5s
+User={{ .User }}
 
 [Install]
 WantedBy=multi-user.target
@@ -194,6 +196,7 @@ func installLinuxService(c *cli.Context) error {
 	}
 	templateArgs := ServiceTemplateArgs{
 		Path: etPath,
+		User: cloudflaredUser,
 	}
 
 	var extraArgsFunc func(c *cli.Context, log *zerolog.Logger) ([]string, error)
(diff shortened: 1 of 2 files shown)


cmd/cloudflared/linux_service.go
```diff
@@ -269,6 +272,13 @@ func installSystemd(templateArgs *ServiceTemplateArgs, log *zerolog.Logger) erro
 			return err
 		}
 	}
+	// Create the cloudflared user if it does not exist
+	if err := runCommand("grep", "-q", fmt.Sprintf("^%s:", cloudflaredUser), "/etc/passwd"); err != nil {
```
sebastianobregoso, line 276: @romeroyonatan probably a best approach would be to use the `id` command.
`id user` would return a non successful exit code if user does not exists, it's probably a better approach than directly reading /etc/passwd.  

On the other hand, `useradd` can be called directly and catch the specific error code if user exists

cmd/cloudflared/linux_service.go
```diff
@@ -339,6 +349,13 @@ func uninstallSystemd(log *zerolog.Logger) error {
 		log.Err(err).Msg("systemctl stop cloudflared-update.timer error")
 		return err
 	}
+	// Delete the cloudflared user if it exists
```
sebastianobregoso, line 352: same

blroot (approved):

01SMD (approved):

crrodriguez:
Nah. 
set

```
DynamicUser=yes
LoadCredential=cloudflared:/etc/cloudflared/<UUID>.json
```
(ensure only root can read that tunnel config file)

and in your config file

` credentials-file: /run/credentials/cloudflared.service/cloudflared`

and voila.. 
if multple tunnels in you can turn the unit file into a template for further flexibility..

01SMD (commented):
.

acuteaura:
Here's all the extra hardening we run.

```
PrivateDevices=true
ProtectControlGroups=true
ProtectHome=true
ProtectKernelTunables=true
ProtectSystem=strict
RestrictSUIDSGID=true
PrivateTmp=true
PrivateUsers=true
NoNewPrivileges=true
CapabilityBoundingSet=
NoExecPaths=/
ExecPaths=/usr/bin/cloudflared
ReadOnlyPaths=/etc/cloudflared
MemoryDenyWriteExecute=true
RemoveIPC=true
RestrictNamespaces=true
```

Issue #734, opened 2022-08-19
🐛 Connect to cloudflared ssh tunnel timeout with vscode


n0k0m3:
**Describe the bug**
Using default generated ssh config + short-lived certs by cloudflared, I was able to ssh using CLI normally. However using vscode to connect to the same hostname will timed out.

**To Reproduce**
0. Have a SSH + Short-lived cert setup on CFZT dashboard
1. Configure ssh config using `cloudflared access ssh-config --hostname <hostname> --short-lived-cert`
2. Use VSCode Remote SSH extension to connect to host using `<username>@<hostname>`
3. Error occurs

**Expected behavior**
VSCode will connect and run normally with cloudflared access tunnel with short-lived certs

**Environment and versions**
 - OS: Linux (Ubuntu Server, Arch Desktop)
 - Architecture: AMD64
 - Version: 2022.8.0

**Logs and errors**
Client side log on vscode
```
[20:20:29.052] Spawned 32316
[20:20:29.132] > local-server-1> Spawned ssh, pid=32324
[20:20:29.134] stderr> OpenSSH_9.0p1, OpenSSL 1.1.1q  5 Jul 2022
[20:20:33.579] stderr> Welcome to Ubuntu 20.04.3 LTS (GNU/Linux 5.13.0-28-generic x86_64)
[20:20:33.579] stderr> 
[20:20:33.579] stderr>  * Documentation:  https://help.ubuntu.com
[20:20:33.580] stderr>  * Management:     https://landscape.canonical.com
[20:20:33.580] stderr>  * Support:        https://ubuntu.com/advantage
[20:20:33.580] stderr> 
[20:20:33.580] stderr>   System information as of Fri 19 Aug 2022 12:20:33 AM UTC
[20:20:33.580] stderr> 
[20:20:33.580] stderr>   System load:                      0.06
[20:20:33.580] stderr>   Usage of /:                       21.1% of 936.92GB
[20:20:33.581] stderr>   Memory usage:                     3%
[20:20:33.581] stderr>   Swap usage:                       0%
[20:20:33.581] stderr>   Temperature:                      53.0 C
[20:20:33.581] stderr>   Processes:                        344
[20:20:33.581] stderr>   Users logged in:                  1
[20:20:33.581] stderr>   IPv4 address for stuffs
[20:20:33.582] stderr> 
[20:20:33.582] stderr>  * Super-optimized for small spaces - read how we shrank the memory
[...]

Issue #574, opened 2022-02-17
Better way to connect with short-lived certs


joey-laminar:
Currently running `cloudflared access ssh-config --hostname mysite.app --short-lived-cert` generates a configuration like this:

```
Host mysite.app
  ProxyCommand bash -c '/usr/local/bin/cloudflared access ssh-gen --hostname %h; ssh -tt %r@cfpipe-mysite.app >&2 <&1'

Host cfpipe-mysite.app
  HostName mysite.app
  ProxyCommand /usr/local/bin/cloudflared access ssh --hostname %h
  IdentityFile ~/.cloudflared/mysite.app-cf_key
  CertificateFile ~/.cloudflared/mysite.app-cf_key-cert.pub
```

The problem with this configuration is that it break scp (which instead opens a shell), as well as any other flags passed to ssh.

A better solution would be to generate a config like this:

```
Match host mysite.app exec "/usr/local/bin/cloudflared access ssh-gen --hostname %h"
  ProxyCommand /usr/local/bin/cloudflared access ssh --hostname %h
  IdentityFile ~/.cloudflared/mysite.app-cf_key
  CertificateFile ~/.cloudflared/mysite.app-cf_key-cert.pub
```

Much simpler and works exactly as expected

cucxabong:
I think this should be a valid option, we're facing this problem also

cucxabong:
Thanks @LoboHacks, it just works :D

nyuszika7h:
how do you give an alias to the host with this configuration? `Match host` only seems to accept FQDNs so `Match host example,ssh.example.com` doesn't work, and this doesn't trigger the `Match host`:
```
Host example
  Host ssh.example.com
```

PR #763, opened 2022-09-22
Issue #574: Better ssh config for short-lived cert

n0k0m3:
This PR is made using suggestion from #574. The pros for this config is that it will work both Windows and Linux (tested), as well as in VSCode, which normally can't be done with the current generated ssh config (refers to #734)


jroyal:
LGTM

PR #766, opened 2022-09-29
Add ingress.Rule Rewrite

moofMonkey:
This PR is inspired by https://github.com/cloudflare/cloudflared/compare/master...regnaio:regnaio/location from https://github.com/cloudflare/cloudflared/issues/563#issuecomment-1081779878.
It adds optional ingress.rule.rewrite parameter, which acts as a replacement string for ingress.rule.path regex.

Slightly modified example from the mentioned issue, which will work with this PR:
```yaml
  - hostname: subdomain.domain.com
    path: ^/subdir1/(.*)$
    service: https://localhost:3001
    rewrite: /$1
  - hostname: subdomain.domain.com
    path: ^/subdir2/(.*)$
    service: https://localhost:3002
    rewrite: /$1
  - hostname: subdomain.domain.com
    path: ^/subdir3/(.*)$
    service: https://localhost:3003
    rewrite: /$1
```

I'm not sure if there is a better way to do this, but reusing existing path regex sounds reasonable to me.
I'm also not sure how to better phrase the comment on ingress.Rule Rewrite.

Side effect from this PR: mockAPI now writes `r.URL.Path` instead of `"Created"` to the response body to test new changes.

### DIFF
--- a/config/configuration.go
+++ b/config/configuration.go
@@ -177,11 +177,11 @@ func ValidateUrl(c *cli.Context, allowURLFromArgs bool) (*url.URL, error) {
 }
 
 type UnvalidatedIngressRule struct {
-	Hostname        string              `json:"hostname,omitempty"`
-	Path            string              `json:"path,omitempty"`
-	PathReplacement string              `yaml:"pathReplacement" json:"pathReplacement,omitempty"`
-	Service         string              `json:"service,omitempty"`
-	OriginRequest   OriginRequestConfig `yaml:"originRequest" json:"originRequest"`
+	Hostname      string              `json:"hostname,omitempty"`
+	Path          string              `json:"path,omitempty"`
+	Rewrite       string              `json:"rewrite,omitempty"`
+	Service       string              `json:"service,omitempty"`
+	OriginRequest OriginRequestConfig `yaml:"originRequest" json:"originRequest"`
 }
 
 // OriginRequestConfig is a set of optional fields that users may set to
--- a/config/configuration.go
+++ b/config/configuration.go
@@ -177,10 +177,11 @@ func ValidateUrl(c *cli.Context, allowURLFromArgs bool) (*url.URL, error) {
 }
 
 type UnvalidatedIngressRule struct {
-	Hostname      string              `json:"hostname,omitempty"`
-	Path          string              `json:"path,omitempty"`
-	Service       string              `json:"service,omitempty"`
-	OriginRequest OriginRequestConfig `yaml:"originRequest" json:"originRequest"`
+	Hostname        string              `json:"hostname,omitempty"`
+	Path            string              `json:"path,omitempty"`
+	PathReplacement string              `yaml:"pathReplacement" json:"pathReplacement,omitempty"`
+	Service         string              `json:"service,omitempty"`
+	OriginRequest   OriginRequestConfig `yaml:"originRequest" json:"originRequest"`
 }
 
 // OriginRequestConfig is a set of optional fields that users may set to
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -117,6 +117,8 @@ func (p *Proxy) ProxyHTTP(
 			originProxy,
 			isWebsocket,
 			rule.Config.DisableChunkedEncoding,
+			rule.Path,
+			rule.PathReplacement,
 			logFields,
 		); err != nil {
 			rule, srv := ruleField(p.ingressRules, ruleNum)
@@ -189,11 +191,18 @@ func (p *Proxy) proxyHTTPRequest(
 	httpService ingress.HTTPOriginProxy,
 	isWebsocket bool,
 	disableChunkedEncoding bool,
+	pathRegexp *ingress.Regexp,
+	pathReplacement string,
 	fields logFields,
 ) error {
 	roundTripReq := tr.Request
-	if isWebsocket {
+	if isWebsocket || pathReplacement != "" {
 		roundTripReq = tr.Clone(tr.Request.Context())
+	}
+	if pathReplacement != "" {
+		roundTripReq.URL.Path = pathRegexp.ReplaceAllString(roundTripReq.URL.Path, pathReplacement)
+	}
+	if isWebsocket {
 		roundTripReq.Header.Set("Connection", "Upgrade")
 		roundTripReq.Header.Set("Upgrade", "websocket")
 		roundTripReq.Header.Set("Sec-Websocket-Version", "13")
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -101,6 +101,11 @@ func (p *Proxy) ProxyHTTP(
 	p.logRequest(req, logFields)
 	ruleSpan.SetAttributes(attribute.Int("rule-num", ruleNum))
 	ruleSpan.End()
+
+	if rule.Rewrite != "" {
+		req.URL.Path = rule.Path.ReplaceAllString(req.URL.Path, rule.Rewrite)
+	}
+
 	if err, applied := p.applyIngressMiddleware(rule, req, w); err != nil {
 		if applied {
 			p.log.Error().Msg(err.Error())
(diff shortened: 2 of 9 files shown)


config/configuration.go
```diff
@@ -177,10 +177,11 @@ func ValidateUrl(c *cli.Context, allowURLFromArgs bool) (*url.URL, error) {
 }
 
 type UnvalidatedIngressRule struct {
-	Hostname      string              `json:"hostname,omitempty"`
-	Path          string              `json:"path,omitempty"`
-	Service       string              `json:"service,omitempty"`
-	OriginRequest OriginRequestConfig `yaml:"originRequest" json:"originRequest"`
+	Hostname        string              `json:"hostname,omitempty"`
+	Path            string              `json:"path,omitempty"`
+	PathReplacement string              `yaml:"pathReplacement" json:"pathReplacement,omitempty"`
```
DevinCarr, line 182: Could we rename this to something like `Rewrite`? I think that would make it more clear that a rewrite is happening on the path before the proxied request is made to the origin.

DevinCarr, line 182: Additionally, what are your thoughts on providing struct here that has two fields? Something like:
```go
type RewritePath struct {
    Regex string
    To    string
}
```
This should allow a bit more flexibility and readability in the Path regex versus the RewritePath regex for those that might want that.

proxy/proxy.go
```diff
@@ -189,11 +191,18 @@ func (p *Proxy) proxyHTTPRequest(
 	httpService ingress.HTTPOriginProxy,
 	isWebsocket bool,
 	disableChunkedEncoding bool,
+	pathRegexp *ingress.Regexp,
+	pathReplacement string,
 	fields logFields,
 ) error {
 	roundTripReq := tr.Request
-	if isWebsocket {
+	if isWebsocket || pathReplacement != "" {
```
DevinCarr, line 199: I'm not sure you need this check here. We still need the isWebsocket, but I don't think we need the pathReplacement check as well.

proxy/proxy_test.go
```diff
@@ -384,7 +406,7 @@ type mockAPI struct{}
 
 func (ma mockAPI) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	w.WriteHeader(http.StatusCreated)
-	_, _ = w.Write([]byte("Created"))
+	_, _ = w.Write([]byte(r.URL.Path))
```
DevinCarr, line 409: Great idea!

proxy/proxy_test.go
```diff
@@ -295,6 +295,18 @@ func TestProxyMultipleOrigins(t *testing.T) {
 	defer api.Close()
 
 	unvalidatedIngress := []config.UnvalidatedIngressRule{
+		{
```
DevinCarr, line 298: Can you add two additional tests:
1. An empty regex replacement (regex: "")
2. At least two regex back-reference replacements (will help others see how to use it, in addition to updated docs)

config/configuration.go
moofMonkey: That sounds good, as it will follow naming similar to NGINX. Done.

moofMonkey: At the first thought, it's unnecessary because path is already a regex which could be safely reused for pattern matching, unlike in NGINX location where it has 3 variations (exact match, regex, prefix and suffix).
After thinking for a while, that might be better for cases where you have hundreds of similarly written ingresses with the same rewrite.

Current PR approach:
```yaml
  - hostname: subdomain.domain.com
    path: ^/serviceGroup/service/(.*)$
    rewrite: /$1
    service: https://localhost:3001
```

NGINX style approach:
```yaml
  - hostname: subdomain.domain.com
    path: ^/serviceGroup/service/
    rewrite:
      - regex: ^/\w+/\w+/(.*)$
        to: /$1
    service: https://localhost:3001
```
or
```yaml
  - hostname: subdomain.domain.com
    path: ^/serviceGroup/service/
    rewrite:
      - regex: ^/serviceGroup/service/(.*)$
        to: /$1
    service: https://localhost:3001
```
It is good because it follows NGINX style, however it allows space for unintentional bugs as regexes are duplicated, which means that they're not guaranteed to match the same path.
If you think that's fine - I'll change it to be this way, but I'd keep it as it is now.

proxy/proxy.go
moofMonkey: I wasn't sure either, but at least tests pass and everything including logging seem to work both with and without that check.
I removed request cloning and moved rewriting right after logging in ProxyHTTP, like in @regnaio branch.

proxy/proxy_test.go
moofMonkey: 1) Do you have any suggestions as to how an empty regex is supposed to work? Right now it reports a configuration error because empty regex doesn't match anything by itself. Should it plainly replace the URL path without back-referencing anything? If so, won't that simply allow for misconfiguration with $'s being sent to services?
2) I don't exactly understand what you meant, but I added a test where /service3/_health/func gets remapped to /func/_health. Is that good enough?

config/configuration.go
DevinCarr: I think that makes sense, we can leave it as a string and it also sets a pretty good boundary. We don't want cloudflared to become a full reverse proxy with all the middleware, and extras. But having this simple rewrite on the path seems like a good middle-ground.

proxy/proxy_test.go
DevinCarr: 1. For the case of the empty regex, I think we want to just skip it. So this would mean that `Path` _must_ exist (and be a valid regex) if you provide a `Rewrite` replacement. If someone wants to provide an empty rewrite from something like `example.com/path` to `example.com` maybe we just recommend that they set the rewrite to `/`? So for the test, just make sure to have a test for `Path: ""` and `Rewrite: "/$1"` (which is expected to fail).
2. Yup, perfect!

moofMonkey: Such a test is already present and specifying rewrite with an empty path is being handled as an error.
This behavior can be changed to silently omit rewrite, but that doesn't seem right.
One thing that could be done is allowing for `Rewrite` to pass with an empty `Path` in case it doesn't contain back-reference replacements, but I don't know whether it's a good idea or not.
https://github.com/cloudflare/cloudflared/blob/4b63132e770f0e633322fb2319f2f2bd6fbf903c/ingress/ingress.go#L297-L299
https://github.com/cloudflare/cloudflared/blob/4b63132e770f0e633322fb2319f2f2bd6fbf903c/ingress/ingress_test.go#L472-L482

Yes, rewriting `example.com/path` to `example.com` is achieved by rewriting to `/`, I don't know how it could be done otherwise without breaking compatibility for old configs while being more or less intuitive.
Given previous example:
```yaml
  - hostname: subdomain.domain.com
    path: ^/path$
    rewrite: /
    service: https://localhost:3001
```
or if you want to rewrite it as a directory:
```yaml
  - hostname: subdomain.domain.com
    path: ^/path/(.*)$
    rewrite: /$1
    service: https://localhost:3001
```

moofMonkey:
`TestWriteToDstSessionPreventClosed` seem to be flaky while other tests pass.
It's probably related to the CI CPU limits and timeout being set to just 100ms in code, it probably needs to be increased to something like 300ms+, but it's out of scope for this PR.

DevinCarr:
First I want to start with that I appreciate all the effort you have put into this PR. After internal discussion, we still want to keep cloudflared lean on the reverse proxy functionality; like rewriting URLs. Going forward, and for the forseable future, our recommendation is to achieve this functionality outside of cloudflared.

You can perform this "middleware" type operations with common reverse proxy software like nginx or caddy. Or, if you prefer to do this with Cloudflare, you can achieve similar results with page rewriting: https://developers.cloudflare.com/rules/transform/url-rewrite/

Again, thank you for all the effort, but I will close this PR. Going forward, I personally will make sure that we (the Cloudflare tunnel team) are more transparent on our scope for cloudflared before proposing and collaborating on outside PRs as to hopefully not guide futile efforts.

Issue #786, opened 2022-10-19
🐛 CVE-2022-32149 found in 2022.10.2


jacobwoffenden:
**Describe the bug**
CVE found:

> Package: golang.org/x/text
> Installed Version: v0.3.7
> Vulnerability CVE-2022-32149
> Severity: MEDIUM
> Fixed Version: 0.3.8
> Link: [CVE-2022-32149](https://avd.aquasec.com/nvd/cve-2022-32149)

**To Reproduce**
Build and scan container using Trivy

**Expected behavior**
Trivy scan passes. golang.org/x/text updated to 0.3.8

**Environment and versions**
GitHub Actions ([link](https://github.com/jacobwoffenden/container-doh-proxy/security/code-scanning/56))

**Logs and errors**
N/A

**Additional context**
N/A

jacobwoffenden:
Closing as this has been marked as resolved https://github.com/jacobwoffenden/container-doh-proxy/security/code-scanning/56

PR #787, opened 2022-10-19
Update golang.org/x/text

jacobwoffenden:
Related to #786


jacobwoffenden:
docker run --rm \
  --volume /var/run/docker.sock:/var/run/docker.sock \
  ghcr.io/aquasecurity/trivy:latest image docker.io/cloudflare/cloudflared-linux-amd64:b1de2a74

PR #801, opened 2022-10-31
Added automatic systemd restart given binary changes

### LINKED ISSUE #799: 🐛Package manager installation: systemd service not being restarted after update
If cloudflared is updated trough package manager, and it is installed as a systemd service, there is no mechanism to restart the service after the update.

**To Reproduce**
1. Install cloudflared from package manager
2. Install cloudflared as service `cloudflared service install`, `systemctl start cloudflared
`
3. Update the package
4. Observe that the service is never restarted, thus the old version remains running until reboot, etc.

**Expected behavior**
There should be a mechanism to restart the service whenever an update is performed.

**Environment and versions**
 - OS: Linux
 - Architecture: AMD
 - Version: 2022.10.3

blroot:
Added a systemd [Path](https://www.freedesktop.org/software/systemd/man/systemd.path.html) that observes changes in the binary and restarts the cloudflared systemd service in that case.

Closes #799

### DIFF
--- a/cmd/cloudflared/linux_service.go
+++ b/cmd/cloudflared/linux_service.go
@@ -71,7 +71,7 @@ Description=Update cloudflared
 After=network.target
 
 [Service]
-ExecStart=/bin/bash -c '{{ .Path }} update; code=$?; if [ $code -eq 11 ]; then systemctl restart cloudflared; exit 0; fi; exit $code'
+ExecStart=/bin/bash -c '{{ .Path }} update'
 `,
 	},
 	{
--- a/cmd/cloudflared/linux_service.go
+++ b/cmd/cloudflared/linux_service.go
@@ -293,6 +293,10 @@ func installSystemd(templateArgs *ServiceTemplateArgs, log *zerolog.Logger) erro
 		log.Err(err).Msg("systemctl enable cloudflared-restart.path error")
 		return err
 	}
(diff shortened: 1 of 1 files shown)


cmd/cloudflared/linux_service.go
```diff
@@ -71,7 +71,7 @@ Description=Update cloudflared
 After=network.target
 
 [Service]
-ExecStart=/bin/bash -c '{{ .Path }} update; code=$?; if [ $code -eq 11 ]; then systemctl restart cloudflared; exit 0; fi; exit $code'
```
blroot, line 74: With these changes, I don't think this explicit restart is useful anymore.

sudarshan-reddy:
What you are asking/doing here is reasonable. However, at this point we feel it best to defer this until:

1. More people want this. We want to be aware of keeping the status quo of the existing behaviour w.r.t service install for now and because point 2 is a workaround. 

2. It is well possible to edit systemd files outside of cloudflared to have this behaviour. As it stands today, you could simply edit the systemd files to give you this feature without having this as part of cloudflared. 

That said, I'd re-iterate that this is a reasonable suggestion and something we could plan for eventually and would be happy to revisit/reopen this PR at a later state.

blroot:
@sudarshan-reddy In case you need to release a security update, do you have a way to force restarting this kind of installations?

PR #870, opened 2023-01-10
GH-352: Add Tunnel CLI option "edge-bind-address"

### LINKED ISSUE #352: [Feature Request] Bind outgoing Argo Tunnel connection to interface or source IP
I have multiple servers with (moderately) complex network setup, and some interfaces have better network connectivity to Cloudflare edge than others. I would like to make `cloudflared` connect to Cloudflare via some preferred interfaces.

Identical feature from other software include:

- Wget `--bind-address=<ip>`
- cURL `curl --interface <ip>`
- `ping -I <interface>` or `ping -I <ip>`
- `mtr -a <ip>`
- `ssh -o BindAddress=<ip>`

I hope `cloudflared` will support something like `--tunnel-bind-ip` or `--tunnel-bind-interface` that will make outgoing connections match the routing policy rules like `from <ip> table <table>` and `from all oif <if> table <table>`.

iBug:
Implements and resolves #352.

- Add the new optional CLI option `--edge-bind-address` for the `tunnel` sub-command.
- Parses into `net.IP` internally.
- New parameter `localAddr net.IP` added to relevant functions, placed immediately after `edgeAddr` (if exists) or before supplementary objects (`logger` etc.)
- At leaf functions, `nil` falls back to `IPv4zero` if required.

Unfortunately I have no idea how to test this the idiomatic way, so I need assistance with this.

My local integration testing shows that the code works as expected:

1. I stop `cloudflared.service`, copy the `ExecStart=` command line from `/etc/systemd/system/cloudflared.service`, replace `argv[0]` with my development compilation, and add `--edge-bind-address <something>`. The final command line is like:

   ```shell
   ~/dev/cloudflared --edge-ip-version 4 --edge-bind-address <something> --protocol auto --no-autoupdate tunnel run --token <redacted>
   ```

2. I open another shell window, run `conntrack -L | grep -F 198.41`, and can see lines like

   ```shell
   tcp      6 431997 ESTABLISHED src=<something> dst=198.41.192.227 sport=55817 dport=7844
   ```

3. I stop the previous `cloudflared` command, replace `<something>` with `<sth_else>` and start again.
4. I run the `conntrack` command again, and can see lines like

   ```shell
   tcp      6 431998 ESTABLISHED src=<sth_else> dst=198.41.200.33 sport=45437 dport=7844
   ```

5. I repeat steps 3-4 with two more IP addresses (connected to different ISPs) on my test machine, and see expected results. Therefore I conclude that my modified code functions properly.
6. Additionally, in case of an invalid argument (`--edge-bind-address asdfghjkl`), I am able to produce this error message:

   ```shell
   ERR Couldn't start tunnel error="invalid value for edge-bind-address: asdfghjkl"
   ```

   In case of a valid IP that does not belong to the host (e.g. `--edge-bind-address 192.0.2.0`), the underlying system call `bind` returns 99 (`EADDRNOTAVAIL`: Cannot assign requested address), and it retries indefinitely.

   ```shell
   ERR Serve tunnel error error="DialContext error: dial tcp 192.0.2.0:0->198.41.200.33:7844: bind: cannot assign requested address" connIndex=0 ip=198.41.200.33
   ```


cmd/cloudflared/tunnel/cmd.go
```diff
@@ -83,12 +83,12 @@ const (
 	LogFieldTmpTraceFilename    = "tmpTraceFilename"
 	LogFieldTraceOutputFilepath = "traceOutputFilepath"
 
-	tunnelCmdErrorMessage = `You did not specify any valid additional argument to the cloudflared tunnel command.
```
iBug, line 86: This one is unintentional but (presumably) harmless: My editor (VS Code) strips trailing whitespaces for me.

DevinCarr:
I like what this change brings to the table. A few things I would like to see before we can continue:
1. `cloudflared` during startup makes a call in edgediscovery to `region1.v2.argotunnel.com` and `region2.v2.argotunnel.com` to get edge addresses that it will attempt to connect to. Some of those IPs can be IPv4 and IPv6 addresses. This change should also override the specified `--edge-ip-version` flag (if provided) as we don't want an error from cloudflared attempting to connect to a IPv4 edge address with a `--edge-bind-address` being IPv6.
2. For the case of "a valid IP that does not belong to the host", could you add an earlier check that validates that the IP is available to the system?

iBug:
@DevinCarr Thanks for the suggestions. I've implemented 1 and it now prints a warning for mismatching `edge-ip-version` and `edge-bind-address`. But for 2 I have some extra thoughts.

- In cb0201de6e8ee527455393c8838a1c8748aa25f2 I implemented that the intuitive way: List all addresses associated with an interface on the host, and find if the supplied `edge-bind-address` is one of them.
- However, there's one caveat here: Linux does not strictly require the address to be on an interface. In fact the necessary and sufficient rule is that the address is routed to the local host\* (e.g. `ip route add local <ip> table local`). Without going platform-specific, the only way to find it out is to actually do a `bind` and see if it errors. But Go does not expose the `bind` call, so in 4879fff29287853008451c94ba471783e98f3ba1 I implemented it using `Dial` towards an [IANA-unassigned UDP port](https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers). I find this better as "connecting" to a UDP service does not produce any traffic (there's no handshakes).

<sup>\* [Ref](https://serverfault.com/a/262334/450575). Even though it's written for IPv6, it is in fact supported for IPv4 as well.</sup>

Let me know which approach you find more appropriate and elegant.

cmd/cloudflared/tunnel/configuration.go
```diff
@@ -348,6 +348,18 @@ func prepareTunnelConfig(
 	if err != nil {
 		return nil, nil, err
 	}
+	edgeBindAddr, err := parseConfigBindAddress(c.String("edge-bind-address"))
+	if err != nil {
+		return nil, nil, err
+	}
+	if err := testIPBindable(edgeBindAddr); err != nil {
```
DevinCarr, line 355: This `edgeBindAddr` can be `nil` as returned by `parseConfigBindAddress`. Should probably be checked before attempting to test the provided IP address.

cmd/cloudflared/tunnel/configuration.go
```diff
@@ -463,6 +476,53 @@ func parseConfigIPVersion(version string) (v allregions.ConfigIPVersion, err err
...
+
+func testIPBindable(ip net.IP) error {
+	var network, address string
+	if ip.To4() != nil {
+		network, address = "udp4", "127.0.0.1:4"
+	} else {
+		network, address = "udp6", "[::1]:4"
+	}
+
+	dialer := net.Dialer{LocalAddr: &net.UDPAddr{IP: ip}}
```
DevinCarr, line 499: Instead of a dialer, since cloudflared will be listening to this address, you can perform this test by listening to this address via `net.ListenUDP`. It also helps reduce the need to specify a port to attempt to dial to since we can let the OS determine which port to listen to.
Something along the lines of this: 
```go
addr, err := net.ResolveUDPAddr("udp", "[::1]:0") // provided edge-bind-address
listener, err := net.ListenUDP("udp", addr)
listener.Close()
```

cmd/cloudflared/tunnel/configuration.go
iBug: Updated `testIPBindable` to simply accept `nil`.

iBug:
Note on the new function `TestTestIPBindable`: It passes when I copy this function into a separate file, but due to the first line of `configuration_test.go` being `//go:build ignore`, the whole file is ignored by Go.

There are other errors in that test file. E.g. `loadOriginCertPool` is undefined - though I found it in `tlsconfig/certreloader.go`, its signature has been changed and I have not looked into this part of the project, so I decided not to mess around.

DevinCarr:
That's fine, I'll take a look at that test file independently of this PR to figure out why we don't build it (likely forgot about it).

cmd/cloudflared/tunnel/configuration.go
```diff
@@ -348,6 +348,18 @@ func prepareTunnelConfig(
...
+	if err != nil {
+		return nil, nil, err
+	}
+	if err := testIPBindable(edgeBindAddr); err != nil {
+		return nil, nil, fmt.Errorf("invalid edge-bind-address %s: %v", edgeBindAddr, err)
+	}
+	edgeIPVersion, err = adjustIPVersionByBindAddress(edgeIPVersion, edgeBindAddr)
+	if err != nil {
+		// This is not a fatal error, we just overrode edgeIPVersion
+		log.Warn().Msgf("Overriding edge-ip-version to %s: %v", edgeIPVersion, err)
```
sssilver, line 361: We generally try to have static messages in log, and embed dynamic parts of the log in structured log fields. In that spirit, this message could have said "Overriding edge-ip-version", and then "edgeIPVersion": edgeIPVersion.

This comment also applies to anywhere else where logging is added.

iBug: Looks like this is the only logging line added in this PR.

iBug:
Local testing shows another possibly missed case: If the user supplies `--edge-bind-address 127.0.0.1`, it will pass the preliminary check but still fails to connect to Cloudflare, with error `sendto: invalid argument`.

However, I'd argue that this is not an actual issue and should not be fixed, for these two reasons:

- Users should be aware of the consequences of supplying `127.0.0.1` or `::1`, compared to mistyping and ending up with a non-local address.
- Explicitly disallowing loopback addresses may interfere with testing, for example if an engineer sets up a mocking CF Edge server and wants to test locally.

DevinCarr:
I think that's fair. We typically offload some level of responsibility to the user whenever an argument is providing an override to something default.

DevinCarr (approved):

joliveirinha (approved):

sssilver:
Please make sure to squash the commits before merging

iBug:
@sssilver Can't this be done via GitHub's "Squash and merge" option?

iBug:
@sssilver Is there any more action required on my side? Or do I just wait for merging?

joliveirinha:
@iBug thanks for the PR.
Everything looks good. We will merge it on our next meeting of public PRs review.

DevinCarr:
@iBug Would you be able to resolve conflicts and then we can merge this!

iBug:
> @iBug Would you be able to resolve conflicts and then we can merge this!

Done.

PR #916, opened 2023-03-15
Add suport for OpenBSD

pufferffish:
I've tested this on my OpenBSD 7.2 machine and this simple fix makes it compile and run flawlessly.


joliveirinha (approved):

01SMD (commented):
.

Issue #923, opened 2023-03-27
🐛 ssh-config w/ short-lived-cert thrown off by invalid characters in hostname


jeslinmx:
**Describe the bug**
The config generated by `cloudflared access ssh-config --short-lived-cert` does not work when `--hostname` contains invalid characters. This prevents using the config directly for otherwise functional addresses like `subdomain.example.com/ssh`.

**To Reproduce**
1. Create tunnel in dashboard, set up `cloudflared tunnel` on SSH server side, etc.
2. Run `cloudflared access ssh-config --hostname subdomain.example.com/ssh --short-lived-cert` on client
  The following config is generated:
  ```
  Add to your /.ssh/config:
  Match host subdomain.example.com/ssh exec "cloudflared access ssh-gen --hostname %h"
    ProxyCommand cloudflared access ssh --hostname %h
    IdentityFile ~/.cloudflared/%h-cf_key
    CertificateFile ~/.cloudflared/%h-cf_key-cert.pub
  ```
  Note the use of `%h` in IdentityFile and CertificateFile
3. Run `ssh subdomain.example.com/ssh`
  This creates the files `subdomain.example.com-ssh-{cf_key|cf_key.pub|cf_key-cert.pub}`. Note that the forward slash has been converted to a hyphen.
4. `ssh` fails with the following error:
  ```
  no such identity: <$HOME>/.cloudflared/subdomain.example.com/ssh-cf_key: No such file or directory
  <$USER>@subdomain.example.com/ssh: Permission denied (publickey).
  ```
  This is as expected, as `%h` = `subdomain.example.com/ssh` != `subdomain.example.com-ssh`.

**Expected behavior**
The config should point to the correct file, such that ssh is able to find the identity file and connect successfully. I propose that `%h` be substituted for the escaped value for the IdentityFile and CertificateFile definitions (optionally, we could instead do this only when the hostname is found to contain invalid characters):

```
Add to your /.ssh/config:
Match host subdomain.example.com/ssh exec "cloudflared access ssh-gen --hostname %h"
  ProxyCommand cloudflared access ssh --hostname %h
  IdentityFile ~/.cloudflared/subdomain.example.com-ssh-cf_key             # replace here
  CertificateFile ~/.cloudflared/subdomain.example.com-ssh-cf_key-cert.pub # and here
```

This is my current workaround, and I can confirm that ssh access works fine in this case.

**Environment and versions**
 - OS: Windows 10 22H2 (build 19045.2604)
[...]

PR #930, opened 2023-04-04
Fix ssh-config short-lived-cert for subpath-ed hostnames

jeslinmx:
Fixes https://github.com/cloudflare/cloudflared/issues/923

### DIFF
--- a/cmd/cloudflared/access/cmd.go
+++ b/cmd/cloudflared/access/cmd.go
@@ -40,8 +40,8 @@ Add to your {{.Home}}/.ssh/config:
 {{- if .ShortLivedCerts}}
 Match host {{.Hostname}} exec "{{.Cloudflared}} access ssh-gen --hostname %h"
   ProxyCommand {{.Cloudflared}} access ssh --hostname %h
-  IdentityFile ~/.cloudflared/%h-cf_key
-  CertificateFile ~/.cloudflared/%h-cf_key-cert.pub
+  IdentityFile ~/.cloudflared/{{.SSHCertFilePath}}-cf_key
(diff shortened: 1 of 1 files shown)


01SMD (approved):

01SMD (approved):

ajvpot:
Is this subject to a race condition when configuring a host name like "*.mycompany.com"? I think connecting to two hosts at the same time could cause undefined behavior.

jeslinmx:
Sorry, but I don't think I understand the potential problem you're describing. Could you give an example of 2 hostnames that might create such a clash?

ajvpot:
Sure, In this case I have configured access like so:

```
alex@Alexs-MacBook-Pro> cloudflared access ssh-config --hostname '*.mycorp.com' --short-lived-cert
Add to your /Users/alex/.ssh/config:

Match host *.mycorp.com exec "/Users/alex/bin/cloudflared access ssh-gen --hostname %h"
  ProxyCommand /Users/alex/bin/cloudflared access ssh --hostname %h
  IdentityFile ~/.cloudflared/%h-cf_key
  CertificateFile ~/.cloudflared/%h-cf_key-cert.pub
```

After this change the config would be:
```
Match host *.mycorp.com exec "/Users/alex/bin/cloudflared access ssh-gen --hostname %h"
  ProxyCommand /Users/alex/bin/cloudflared access ssh --hostname %h
  IdentityFile ~/.cloudflared/*.mycorp.com-cf_key
  CertificateFile ~/.cloudflared/*.mycorp.com-cf_key-cert.pub
```

Which may cause problems when SSHing to multiple hosts that match that pattern and use different SSH CA public keys at the same time.

PR #1020, opened 2023-07-11
http2 tunnel support socks5 proxy from env all_proxy

### LINKED ISSUE #350: Cloudflared tunnel via proxy
I see that `cloudflared access` has support for a proxy #317.

I'm trying to use `cloudflared tunnel` as well behind a corporate firewall, but this doesn't seem to support proxy? The connection attempt is always direct.

Asutorufa:
close #350 #1025

### DIFF
--- a/edgediscovery/dial.go
+++ b/edgediscovery/dial.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
+	"golang.org/x/net/proxy"
 )
 
 // DialEdgeWithH2Mux makes a TLS connection to a Cloudflare edge node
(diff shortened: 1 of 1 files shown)


ddvnguyen:
Recently I want to replace ngrok by cloudflared tunnel.
When come to fullstack development, we want to track https requests we get content in requests and also able to replay requests.
All of that things are supported out of the box by ngrok. So when using cloudflared tunnel I plan to use mitmproxy together so it can do more things with all http requerst coming to the dev servers.
I have been setup a lot of ways still can not make it work. I also try to use "proxychains" but seem like some how the request did not go though the proxy server.
Thank to this change request now it works.

For someone who interesting in how I config Cloudflare tunnel and mitmproxy:
Current my solution is to build a Ubuntu docker image that install both mitmproxy and Cloudflare.

For mitmproxy I use command: mitmweb --listen-port 9099 --listen-host 0.0.0.0 --web-port 9091 --web-host 0.0.0.0
this command allows access web ui in my windows host machine also mitmproxy will call directly to host we don't need to expso to many port as we config more dev servers later.

For Cloudflare tunnel if you want to tunnel a localhost in host machine like me, you need to replace "localhost" with "host.docker.internal". just make sure set proxy in the container.

Hope that this merge request will be accepted soon.

SeaX:
Hi @joliveirinha, is there anyway to use Cloudfare Tunnels in a corporate environment when in the vast majority of cases there is a need of proxies ?

dong-dong6:
> 你好@joliveirinha，在绝大多数情况下需要代理的情况下，有没有办法在企业环境中使用 Cloudfare Tunnels？
> 
> 为了建立隧道，我们需要以下 Cloudflared ------> HTTP 代理 ------> Cloudflare

I also need this function to speed up the document cloudflaretunnel.

PR #1026, opened 2023-07-17
Write tunnel token to environment file

### LINKED ISSUE #802: 💡Mask token in argv when using the `--token` flag
**Describe the feature you'd like**
I know that a config file can be used for this, But I think a secret should not be visible in the command line when running some tools like `ps`, `htop`, or etc.

iBug:
Currently implemented for systemd service, fixes #802


chungthuang:
Hi @iBug, thank you for taking the time to look into this. However, I think you can achieve the same effect by changing systemd service template after `cloudflared service install`.

iBug:
@chungthuang So why not make this the default behavior for all users?

sixcorners:
@joliveirinha @chungthuang If you don't want to change the software to be secure by default and instead want us to modify the config after it installs the insecure version could you document the steps on how to not expose the token on the process's command line?

Issue #888, opened 2023-02-04
FTBFS due to quic-go not supporting go 1.20


grawlinson:
**Describe the bug**
cloudflared does not compile on go 1.20 due to the quic-go dependency specified in go.mod not supporting 1.20.

As of ~13 hours ago, 1.20 is [supported](https://github.com/quic-go/quic-go/commit/5b72f4c900f209b5705bb0959399d59e495a2c6e) upstream.

**To Reproduce**
Compile with go 1.20 and it will fail with the following error:

```sh
# github.com/lucas-clemente/quic-go/internal/qtls
../pkg/mod/github.com/chungthuang/quic-go@v0.27.1-0.20220809135021-ca330f1dec9f/internal/qtls/go120.go:6:13: cannot use "The version of quic-go you're using can't be built on Go 1.20 yet. For more details, please see https://github.com/lucas-clemente/quic-go/wiki/quic-go-and-Go-versions." (untyped string constant "The version of quic-go you're using can't be built on Go 1.20 yet. F...) as int value in variable declarat
```

The dependency on quic-go should probably be updated sooner rather than later.

sudarshan-reddy:
Thanks for flagging this @grawlinson

mraerino:
this is going to be extra tricky, since cloudflared does not use upstream quic-go, but a fork.
would you accept a PR that changed to using the latest upstream quic-go release?

bwesterb:
We actually do use upstream quic-go, but replace the underlying qtls library. [I prepared one for 1.20](https://github.com/cloudflare/qtls-pq/tree/go1.20) yesterday, so it's ready to go.

b0gdanw:
cloudflared-2023.3.1
go version go1.20.2 darwin/arm64

# github.com/lucas-clemente/quic-go/internal/qtls
vendor/github.com/lucas-clemente/quic-go/internal/qtls/go120.go:6:13: cannot use "The version of quic-go you're using can't be built on Go 1.20 yet. For more details, please see https://github.com/lucas-clemente/quic-go/wiki/quic-go-and-Go-versions." (untyped string constant "The version of quic-go you're using can't be built on Go 1.20 yet. F...) as int value in variable declaration

stefanb:
FYI, go 1.21 will be released soon (see [rc2 announcement](https://groups.google.com/g/golang-announce/c/REcCdQj1taA)), which will make go 1.19 not maintained anymore.

AndrewBucklin:
[...]

PR #1027, opened 2023-07-18
Run checks with Go 1.20

stefanb:
Go 1.19 is unmaintained since Go 1.21 was released.

Go 1.20 compatibility was added since v2023.5.1 via [TUN-7227: Migrate to devincarr/quic-go](https://github.com/cloudflare/cloudflared/commit/9426b603082905d0af8a07bdac866bc1d9c37cba)

See also 
* #1047
* #888 
* a more thorough replacement PR: #1052
* needs further work for Go 1.21: #1043

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -4,7 +4,7 @@ jobs:
   check:
     strategy:
       matrix:
-        go-version: [1.19.x]
+        go-version: [1.20.x]
         os: [ubuntu-latest, macos-latest, windows-latest]
     runs-on: ${{ matrix.os }}
     steps:


.github/workflows/check.yaml
```diff
@@ -4,7 +4,7 @@ jobs:
   check:
     strategy:
       matrix:
-        go-version: [1.19.x]
+        go-version: [1.20.x]
```
stefanb, line 7: Once this workflow starts using `actions/setup-go@v4` instead of `v3` it could be even simplified:
```suggestion
        go-version: ['stable']
```
...but that could have undesired consequences, breaking the checks once a new `stable` go version is released.

stefanb:
Done in [TUN-7584: Bump go 1.20.6](https://github.com/cloudflare/cloudflared/commit/65247b6f0f2f354eee78d9b3e3d987dff59442a8).

PR #1030, opened 2023-07-20
Fix config update when config is "null"

### LINKED ISSUE #1029: 🐛
**Describe the bug**
The config provided via `--config` is ignored since version 2023.6.0 - everything works as expected up until 2023.5.0.

**To Reproduce**
1. Create a tunnel in Cloudflare Zero Trust
2. Create a valid config.yaml with an ingress rule
3. Run `docker run -v ./config.yaml:/etc/cloudflared-config.yaml cloudflare/cloudflared:latest tunnel --no-autoupdate --config /etc/cloudflared-config.yaml run --token $TOKEN`

If it's an issue with Cloudflare Tunnel:
- Tunnel ID: 233a437b-b8d5-4b87-9b8f-e4bf76f6d4ab
- cloudflared config: any valid config with an ingress rule

When you run the above you will see in the logs:

    INF Updated to new configuration config=null version=0
    ...
    WRN No ingress rules were defined in provided config (if any) nor from the cli, cloudflared will return 503 for all incoming HTTP requests

**Expected behavior**
If the config is provided with `--config` cloudflared should not overwrite the config with the one from the cloud.

**Environment and versions**
 - OS: any
 - Architecture: any
 - Version: 2023.7.1

**Logs and errors**
WRN No ingress rules were defined in provided config (if any) nor from the cli, cloudflared will return 503 for all incoming HTTP requests

**Additional context**
2023.6.0 seems to be the first version that has this bug.

derkoe:
When the tunnel reads the config from the server and it is "null" then ignore it.

Fixes #1029

### DIFF
--- a/orchestration/orchestrator.go
+++ b/orchestration/orchestrator.go
@@ -74,6 +74,15 @@ func (o *Orchestrator) UpdateConfig(version int32, config []byte) *tunnelpogs.Up
 			LastAppliedVersion: o.currentVersion,
 		}
 	}
+
+	if string(config) == "null" {
+		o.log.Info().Msg("Ignore null config from server")
+		o.currentVersion = version
+		return &tunnelpogs.UpdateConfigurationResponse{
+			LastAppliedVersion: o.currentVersion,
+		}
+	}
+
 	var newConf newRemoteConfig
 	if err := json.Unmarshal(config, &newConf); err != nil {
 		o.log.Err(err).


derkoe:
This only partially fixes #1029 - the config should not be loaded from the server when `--config` is specified.

joliveirinha:
Answered on the ticket itself for the reason why we don't want to allow this.

PR #1040, opened 2023-08-08
[access] use os.Executable to discover the path to cloudflared when generating ssh config

ajvpot:
Uses os.Executable to discover the path to cloudflared when generating ssh config. If os.Executable is unavailable, falls back to the previous method.

This allows for more accurate generation of the ssh config. Previously, if cloudflared was installed in the PATH, an incorrect config could be generated. This also allows for inclusion of the subcommand in other `cli/v2` apps that are not named `cloudflared`.

image


jroyal (approved):

ajvpot:
Hey @jroyal is there anything I can do to move this PR ahead? Thanks!

joliveirinha (approved):

Issue #843, opened 2022-12-09
💡Option to disable remote configuration of a Tunnel


AlbertSPedersen:
When cloudflared establishes a connection to Cloudflare's edge, the origintunnel service sends the remote configuration to cloudflared (if available), and cloudflared happily applies it without a second thought, overwriting the local configuration.

As I will explain below, this can be a problem in some cases. Therefore I propose adding an option to cloudflared that disables remote configuration of a Tunnel.

### Locally managed Tunnels

With a locally managed Tunnel, WARP-routing is disabled by default and must be explicitly enabled by adding the following keys to the configuration file.

```yaml
warp-routing:
  enabled: true
```

From what I can can tell, having to explicitly enable `warp-routing` is meant as a security feature. In the event of a Cloudflare account compromise, even if an attacker has access to modify the network routing configuration of the ZT organization, they will be unable to turn the Tunnel into a "jump host" for its private network.

### The Problem: Tunnel Migration

The problem is that, from what I can tell, it is not currently possible to tell cloudflared to ignore the remote configuration.

This means that, even if your Tunnel is locally managed and has WARP-routing disabled, in the event of a Cloudflare account compromise (or a vulnerability in Cloudflare), an attacker can simply convert the locally managed Tunnel to a remotely managed Tunnel and then enable WARP-routing.

While Cloudflare account compromise is unlikely if proper security is in place, it should be possible to take steps to prevent lateral movement in the unfortunate event that it does happen.

joliveirinha:
@AlbertSPedersen what do you mean by:
"his means that, even if your Tunnel is locally managed and has WARP-routing disabled, in the event of a Cloudflare account compromise (or a vulnerability in Cloudflare)" ?

if your tunnel is locally managed, then it is not remotely managed. Meaning that our services will not push the configuration to cloudflared.

Are you sure you are creating locally managed tunnels?

jeremydonahue:
I up-thumbed this issue, but also wanted to add that it would be nice if there was a way to convert a remotely managed tunnel back to a locally managed tunnel. All it takes is one well intentioned person to hit the _Configure_ button in the Zero Trust dashboard, and now all local configuration is ignored for the tunnel and the service stops working. The only solution that I know of right now is to delete and recreate the tunnel. We prefer locally managed tunnels for the ability to manage the configuration with Git and pretty much never want remotely managed tunnels to be used.

tjb900:
[...]

PR #1086, opened 2023-10-19
GH-843: Add option to disable remote configuration

yapret:
Implements #843 

- Add a new `--no-remote-config` global CLI option
- When the option is enabled:
    - Stop advertising `FeatureAllowRemoteConfig` in the `clientFeatures` passed to `tunnelpogs`
    - Ignore incoming remote config updates in `Orchestrator`

`cloudflared` seemed to behave correctly while testing locally:
- On a migrated remotely-managed Tunnel, config from the Zero Trust dashboard was ignored, both at startup and when updating it, while the local config kept being enforced
- Migrating an established locally-managed Tunnel to remote management doesn't disturb it, traffic continues to flow as if nothing had happened

For better UX, Cloudflare could implement an additional requirement on the ZT dashboard when `--no-remote-config` is enabled to prevent migration from local to remote management, or more generally a warning banner on the Tunnel configuration page (the option already gets displayed in the UI so I'd guess it's a small change).

Not a Gopher, advice is very welcome!

### DIFF
--- a/cmd/cloudflared/tunnel/cmd.go
+++ b/cmd/cloudflared/tunnel/cmd.go
@@ -782,6 +782,12 @@ func tunnelFlags(shouldHide bool) []cli.Flag {
 			EnvVars: []string{"TUNNEL_MANAGEMENT_DIAGNOSTICS"},
 			Value:   false,
 		}),
+		altsrc.NewBoolFlag(&cli.BoolFlag{
+			Name:    config.DisableRemoteConfigFlag,
+			Usage:   "Disables remote configuration from the edge",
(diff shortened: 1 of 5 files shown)


joliveirinha:
You can already create a tunnel that is locally managed. 
This is not something we want to do, for that reason, we are going to close the PR.

We appreciate your contribution, however, in the future, feel free to create an issue first to request the product team if your idea is aligned with product. If so, we are happy to accept product functionality changes.

tjb900:
@joliveirinha is there any chance you (or the product team) might be able to elaborate more on why this isn't desirable / aligned with the direction of the product? Are locally-managed tunnels effectively being deprecated?

PR #1112, opened 2023-11-19
 Remove 'custom:' from hostnames (label)

flpydsk:
Remove 'custom:' from hostname/label value


DevinCarr:
This would likely break functionality in the UI, we will need to address this internally first if we want to make this change. For now I will decline this and discuss it internally with the UI team.

Issue #1053, opened 2023-08-21
Cloudflared on FreeBSD🐛


kinchan:
Hi
First of all, Great thank you for cloudflared project members.
I used FreeBSD 13.2 now.
FreeBSD have cloudflared ports program, but looks like no mentenansu now.
So, I tried

Install lang/go,gmake.

git clone https://github.com/cloudflare/cloudflared.git

and, gmake but error below.
-------
ok      github.com/cloudflare/cloudflared/websocket     0.059s
FAIL
gmake: *** [Makefile:147: test] Error 1
------

So, I tried Makefile change below comment out.

-------
.PHONY: test
test: vet
#ifndef CI
#        go test -v -mod=vendor -race $(LDFLAGS) ./...
#else
#        @mkdir -p .cover
#        go test -v -mod=vendor -race $(LDFLAGS) -coverprofile=".cover/c.out" ./...
#endif
------

and gmake again.
So, no error, made cloudflared program.

I used it, look like work fine.

Excuse me, please advice me.

Thank you for reading.

PR #1115, opened 2023-11-20
Added FreeBSD builds

robvanoostenrijk:
As per issue #1053, currently no FreeBSD builds are available.

This PR updates the build_packages.sh script to add `cloudflared-freebsd-*` builds.


joliveirinha:
We don't officially support FreeBSD at this time. For this reason, we are not going to accept this PR at this stage.

@abelinkinbio

PR #1123, opened 2023-12-14
cmd/cloudflared: fix dropped error

alrs:
This picks up an `err` variable that was being declared inside an `if` block with the `:=` operator and thus lost before `return`.

### DIFF
--- a/cmd/cloudflared/service_template.go
+++ b/cmd/cloudflared/service_template.go
@@ -124,8 +124,9 @@ func openFile(path string, create bool) (file *os.File, exists bool, err error) 
 	if err != nil {
 		return nil, false, err
 	}
+	var fileInfo os.FileInfo
 	if create {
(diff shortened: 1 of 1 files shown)


cmd/cloudflared/service_template.go
```diff
@@ -124,8 +124,9 @@ func openFile(path string, create bool) (file *os.File, exists bool, err error)
 	if err != nil {
 		return nil, false, err
 	}
+	var fileInfo os.FileInfo
 	if create {
-		fileInfo, err := os.Stat(expandedPath)
+		fileInfo, err = os.Stat(expandedPath)
```
chungthuang, line 129: This is a good catch, but we need to check if this returns an error and return it, otherwise line 133 will override the error.

chungthuang:
Closing due to inactivity.

PR #1130, opened 2023-12-24
fix checkInPingGroup bugs

### LINKED ISSUE #1127: error="failed to determine minimum ping group ID: strconv.ParseInt: parsing \"4294967294\": value out of range
**Describe the bug**

The user running cloudflared process has a GID (group ID) that is not within ping_group_range. You might need to add that user to a group within that range, or instead update the range to encompass a group the user is already in by modifying /proc/sys/net/ipv4/ping_group_range. Otherwise cloudflared will not be able to ping this network error="failed to determine minimum ping group ID: strconv.ParseInt: parsing \"4294967294\": value out of range"

The parser WAS correct until https://github.com/torvalds/linux/commit/e209fee4118fe9a449d4d805361eb2de6796be39 but now it is incorrect as a wider range is allowed and valid.

crrodriguez:
- Must check for the *effective* GID. 
- Must allow range from  0 to 4294967294 in current kernels.
- fixes #1127

### DIFF
--- a/ingress/icmp_linux.go
+++ b/ingress/icmp_linux.go
@@ -78,19 +78,19 @@ func checkInPingGroup() error {
 	if err != nil {
 		return err
 	}
-	groupID := os.Getgid()
+	groupID := uint64(os.Getegid())
 	// Example content: 999	   59999
 	found := findGroupIDRegex.FindAll(file, 2)
 	if len(found) == 2 {
-		groupMin, err := strconv.ParseInt(string(found[0]), 10, 32)
+		groupMin, err := strconv.ParseUint(string(found[0]), 10, 32)
 		if err != nil {
 			return errors.Wrapf(err, "failed to determine minimum ping group ID")
 		}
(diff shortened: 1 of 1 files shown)


ingress/icmp_linux.go
```diff
@@ -78,19 +78,19 @@ func checkInPingGroup() error {
 	if err != nil {
 		return err
 	}
-	groupID := os.Getgid()
+	groupID := uint64(os.Getegid())
```
jcsf, line 81: Can you revert this to use the old method. I don't believe it is required to use the effective group id.
```suggestion
	groupID := uint64(os.Getgid())
```

chungthuang: Hi @crrodriguez is this documented? Can you share a link to it?

crrodriguez: there  --> https://github.com/torvalds/linux/blob/026e680b0a08a62b1d948e5a8ca78700bfac0e6e/net/ipv4/ping.c#L255
it compares the effective group id with the list

chungthuang (approved):

PR #1134, opened 2023-12-28
Do not logs errors, when all possible errors are already returned for others to handle

firecow:


jcsf:
Hi,
Thank you for your contribution, but we can't remove those logs because they provide useful information to the user, which isn't present in the errors. Therefore, the only way to remove them would be add the log message to the errors, which creates very long errors, which doesn't make that much sense. So, it is better to have the error causing 2 different logs over not providing a proper logging for the user to fix the issue.

Issue #686, opened 2022-07-05
🐛 post-removal script is not cleaning everything


eximius313:
**Describe the bug**
Installation is successful, but the purging (`dpkg --purge cloudflared`) does not clean the service
```
$ ls /etc/systemd/system/cloudflared*
cloudflared.service 
cloudflared-update.service  
cloudflared-update.timer
```

nor config files:
```
$ ls /etc/cloudflared/
config.yml
```

therefore purge shows following errors:
```
dpkg: error processing package cloudflared (--purge):
 installed cloudflared package post-removal script subprocess returned error exit status 1
```

and after reinstall (`apt-get install --reinstall cloudflared`) when one tries to install new service, they get:
```
cloudflared service is already installed at /etc/systemd/system/cloudflared.service; if you are running a cloudflared tunnel, you can point it to multiple origins, avoiding the need to run more than one cloudflared service in the same machine; otherwise if you are really sure, you can do `cloudflared service uninstall` to clean up the existing service and then try again this command
```

Moreover - I was getting weird errors while trying to `systemctl disable cloudflared`:
```
Unit /etc/systemd/system/cloudflared.service is masked, ignoring.
```

**To Reproduce**
1. `dpkg --purge cloudflared`

**Expected behavior**
`post-removal script` should clean everything.

**Environment and versions**
 - OS: Linux
 - Architecture: Arm64
 - Version: 2022.6.3

Soulsender:
Had this same issue, this solution worked for me:

```sudo mv /var/lib/dpkg/info/cloudflared.* /tmp/
sudo dpkg --remove --force-remove-reinstreq cloudflared
sudo apt-get remove cloudflared
sudo apt-get autoremove && sudo apt-get autoclean
```

fabiannagel:
+1, still an issue

rubenz91:
Same here, mine kind of crashed, then after doing the official step to uninstall left garbage that dint allow to uninstall and now its even worse as states the follow doing the cleanups: 

`E: The package cloudflared needs to be reinstalled, but I can't find an archive for it.
`

Fir the dpkg:

`sudo dpkg --remove --force-remove-reinstreq cloudflared
dpkg: warning: overriding problem because --force enabled:
dpkg: warning: package is in a very bad inconsistent state; you should
 reinstall it before attempting a removal
(Reading database ... 222495 files and directories currently installed.)
Removing cloudflared (2023.7.3) ...
[...]

PR #1166, opened 2024-01-23
Update postrm.sh to fix incomplete uninstall

NanashiTheNameless:
this should supposedly fix https://github.com/cloudflare/cloudflared/issues/686 if i remember the issue correctly when i faced it, as "-f" will make it not hard fail if the file is not found.

### DIFF
--- a/postrm.sh
+++ b/postrm.sh
@@ -1,4 +1,4 @@
 #!/bin/bash
 set -eu
-rm /usr/local/bin/cloudflared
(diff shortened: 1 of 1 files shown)


b4shful (commented):
This change is in line with other postrm scripts in other projects which conventionally use `rm -f` and not `rm` on its own. Hopefully we can get someone to review and approve! This would fix quite a few instances of the same issue online (outside of GitHub) too :)

jcsf (approved):

Issue #1243, opened 2024-05-10
💡Support ingress rule matching for bastion mode


shayonj:
**Describe the feature you'd like**
Be able to support ingress rules in bastion Mode

**Context**

👋🏾 Hello from [Tines](https://tines.com/)! I work on the Platform Team, and we're big fans and heavy users of CF Tunnel. In a nutshell, we use Tunnel in our multi-tenant environment to proxy our customers' HTTP workloads from our cloud instances to their on-prem/self-hosted instances through Cloudflared Tunnel. We've built a small orchestration layer on top of the Tunnel that dynamically proxies our customers' requests from our HTTP client middleware via `cloudflared access` (entry node) to the customers' network (exit node).

Our setup operates in bastion mode, meaning the proxied request has a single destination.

**Problem**

We are looking to eliminate any possibility of "data exfiltration." As you might guess, the exit node where the `cloudflared` tunnel runs can access virtually any service within its network boundary.

Tunnel Ingress Rules is an excellent feature that allows traffic routing to different services based on hostname. It also enables "blackhole-ing" incoming requests that do not match a certain hostname by declaring `http_status:404` as the service, for example.

However, ingress rules do not work with `bastionMode`. We could avoid using bastion mode, but given the scale (many multi tenant clusters) and dynamism of how we run Tunnels, this would require us to provision a new hostname/DNS per Tunnel in a multi-tenant env to leverage ingress rules in non-bastion mode, further necessitating some UX/DX changes.

**Proposal**

My proposal is to extend the existing functionality of `bastionMode`. If `bastionMode` is set to true, and a `hostname` and `service` name are provided as part of the ingress rules config, then we attempt to match the `hostname` against the header `Cf-Access-Jump-Destination` in the incoming HTTP request. If the `Rule` matches, then we accordingly proxy the bastion request to the mentioned `service`. If a request does not match any service, then it falls into the usual catch-all behavior.

**Describe alternatives you've considered**
Mentioned above

**Additional context**
I wanted to see if this is something thats possible and how much lift would it be. I was able to try this out in our setup and works as expected. So, I took a stab at a PR here: https://github.com/cloudflare/cloudflared/pull/1244

 I would also love to learn if there are opportunities to improve this approach or consider other options.

Given our reliance on Tunnel, we would also be happy to discuss what continued support and maintenance would look like.

Example config

```yaml
tunnel: aaaa-bbbb-cccc

ingress:
  - hostname: test-foobar-service.com
[...]

PR #1244, opened 2024-05-10
Proposal: Support ingress rule matching for bastion mode

shayonj:
## Context

👋🏾 Hello from [Tines](https://tines.com/)! I work on the Platform Team, and we're big fans and heavy users of CF Tunnel. In a nutshell, we use Tunnel in our multi-tenant environment to proxy our customers' HTTP workloads from our cloud instances to their on-prem/self-hosted instances through Cloudflared Tunnel. We've built a small orchestration layer on top of the Tunnel that dynamically proxies our customers' requests from our HTTP client middleware via `cloudflared access` (entry node) to the customers' network (exit node).

Our setup operates in bastion mode, meaning the proxied request has a single destination.

## Problem

We are looking to eliminate any possibility of "data exfiltration." As you might guess, the exit node where the `cloudflared` tunnel runs can access virtually any service within its network boundary.

Tunnel Ingress Rules is an excellent feature that allows traffic routing to different services based on hostname. It also enables "blackhole-ing" incoming requests that do not match a certain hostname by declaring `http_status:404` as the service, for example.

However, ingress rules do not work with `bastionMode`. We could avoid using bastion mode, but given the scale (many multi tenant clusters) and dynamism of how we run Tunnels, this would require us to provision a new hostname/DNS per Tunnel in a multi-tenant env to leverage ingress rules in non-bastion mode, further necessitating some UX/DX changes.

## Proposal

My proposal is to extend the existing functionality of `bastionMode`. If `bastionMode` is set to true, and a `hostname` and `service` name are provided as part of the ingress rules config, then we attempt to match the `hostname` against the header `Cf-Access-Jump-Destination` in the incoming HTTP request. If the `Rule` matches, then we accordingly proxy the bastion request to the mentioned `service`. If a request does not match any service, then it falls into the usual catch-all behavior.

I have tested this setup in our environment, and it works as expected. That said, I'd like to propose this as a feature request for the project. I would also love to learn if there are opportunities to improve this approach or consider other options.

Given our reliance on Tunnel, we would also be happy to discuss what continued support and maintenance would look like.

Example config

```yaml
tunnel: aaaa-bbbb-cccc

ingress:
  - hostname: test-foobar-service.com
    service: http_status:200
  - hostname: "foobar.tunnel-dev.com"
    service: "http://localhost:9000"
  - hostname: "foobar1.tunnel-dev.com"
    service: "http://localhost:9002"
    originRequest:
      bastionMode: true
  - service: http_status:404
``` 

https://github.com/cloudflare/cloudflared/issues/1243

### DIFF
--- a/ingress/ingress.go
+++ b/ingress/ingress.go
@@ -28,7 +28,6 @@ var (
 )
 
 const (
-	ServiceBastion     = "bastion"
 	ServiceSocksProxy  = "socks-proxy"
 	ServiceWarpRouting = "warp-routing"
 )
@@ -38,12 +37,13 @@ const (
 // which is the case if the rules were instantiated via the ingress#Validate method.
 //
 // Negative index rule signifies local cloudflared rules (not-user defined).
-func (ing Ingress) FindMatchingRule(hostname, path string) (*Rule, int) {
+func (ing Ingress) FindMatchingRule(hostname, path string, cfJumpDestinationHeader string) (*Rule, int) {
 	// The hostname might contain port. We only want to compare the host part with the rule
 	host, _, err := net.SplitHostPort(hostname)
 	if err == nil {
 		hostname = host
 	}
+	derivedHostName := hostname
 	for i, rule := range ing.InternalRules {
 		if rule.Matches(hostname, path) {
 			// Local rule matches return a negative rule index to distiguish local rules from user-defined rules in logs
@@ -52,7 +52,15 @@ func (ing Ingress) FindMatchingRule(hostname, path string) (*Rule, int) {
 		}
 	}
 	for i, rule := range ing.Rules {
-		if rule.Matches(hostname, path) {
+		// If bastion mode is turned on and request is made as bastion, attempt
+		// to match a rule where jump destination header matches the hostname
+		if rule.Config.BastionMode && len(cfJumpDestinationHeader) > 0 {
+			jumpDestinationUri, err := url.Parse(cfJumpDestinationHeader)
+			if err == nil {
+				derivedHostName = jumpDestinationUri.Hostname()
+			}
+		}
+		if rule.Matches(derivedHostName, path) {
 			return &rule, i
 		}
 	}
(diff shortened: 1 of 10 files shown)


ingress/ingress.go
```diff
@@ -28,7 +28,6 @@ var (
 )
 
 const (
-	ServiceBastion     = "bastion"
```
shayonj, line 31: Consolidate the use via `config.BastionFlag` and also to avoid `import cycle not allowed in test` which got raised when I was referencing `ingress` from `carrier`

proxy/proxy.go
```diff
@@ -86,55 +87,65 @@ func (p *Proxy) ProxyHTTP(
...
-		}
-		return nil
 	case ingress.HTTPLocalProxy:
 		p.proxyLocalRequest(originProxy, w, req, isWebsocket)
-		return nil
 	default:
 		return fmt.Errorf("Unrecognized service: %s, %t", rule.Service, originProxy)
 	}
+	return nil
 }
```
shayonj, line 149: extracted the logic into `handleHTTPBasedService` and `handleStreamBasedService`

shayonj:
👋🏾 Apologies for the direct ping @chungthuang - I was curious if you were able to share any insight into the timeline around when can I expect some feedback on this? Just trying to decide accordingly if I could wait on a release or proceed with a temporary fork for now. Thank you so much for you work 🙏🏾

shayonj:
@jcsf @DevinCarr if you are able to help take a look at the PR and/or share any insights into the review process that would be great 🙏🏾

janani-cr:
@shayonj Thanks for your contribution, we'll review this PR in our next meeting which will most likely not align with your timeline, hence our recommendation would be for you to do a fork to carry on with your work.

shayonj:
Thank you so much @janani-cr, really appreciate it 🙏🏾

shayonj:
👋🏾 Hi @janani-cr - just curious if you are able to share any insight into by when we can expect a review on the PR? Thank you

shayonj:
👋🏾 Just a friendly ping on the above. Would love to know by when we could receive a review, thank you!

shayonj:
👋🏾 @janani-cr - Would love to know if this is something you are consider for the `main` branch? Or if any feedback you/team has?

Thanks!

jcsf:
As discussed, this doesn't align with the zero trust principles, therefore we will not be accepting this PR.
Regardless, thank your for your contribution and interest in this project.

Issue #1249, opened 2024-05-20
💡option for cloudflared access login to only ouput jwt


rjgpacheco:
**Describe the feature you'd like**
Add -no-pretty flag to cloudflared access login making it only output jwt token: 

```shell
$ cloudflared access login -no-pretty $url
<TOKEN>
```

**Describe alternatives you've considered**
Are there any alternatives to solving this problem? If so, what was your experience with them?

**Additional context**

PR #1250, opened 2024-05-20
feat: option for cloudflared access login to only output jwt (#1249)

rjgpacheco:
Adds a -no-pretty flag to cloudflared access login:

```shell
$ cloudflared access login -no-pretty $url
<TOKEN>
```

This removes the need to parse the token from the output message, making it easier for external applications.

This is a pretty small change change that makes it much simpler to use from, eg Python script.

With the current CLI, the caller must manually parse the output, eg:

```python
from subprocess import check_output

def get_auth_token(url) -> str:
    cmd = ["cloudflared", "access", "login", url]
    return check_output(cmd).decode("utf-8").strip().split()[-1]

```

Python's the simplest case here, but it's a step that doesn't really need to happen.

Also: the hardest part is coming up with a name for the flag, so I didn't thing too much about it, open to suggestions :)

### DIFF
--- a/cmd/cloudflared/access/cmd.go
+++ b/cmd/cloudflared/access/cmd.go
@@ -27,6 +27,7 @@ import (
 
 const (
 	loginQuietFlag     = "quiet"
+	noPrettyFlag       = "no-pretty"
 	sshHostnameFlag    = "hostname"
 	sshDestinationFlag = "destination"
 	sshURLFlag         = "url"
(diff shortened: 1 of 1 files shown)


janrueth:
`--only-token` might be an alternative. But I like this feature, makes using cloudflared in bash scripts a lot simpler

janrueth (commented):
--only-token might be an alternative. But I like this feature, makes using cloudflared in bash scripts a lot simpler

PR #1289, opened 2024-07-08
Use RELEASE_NOTES date instead of build date

bmwiedemann:
Use `RELEASE_NOTES` date instead of build date
to make builds reproducible.
See https://reproducible-builds.org/ for why this is good
and https://reproducible-builds.org/specs/source-date-epoch/
for the definition of this variable.
This date call only works with GNU date and BSD date.

Alternatively, https://reproducible-builds.org/docs/source-date-epoch/#makefile could be implemented.

This patch was done while working on reproducible builds for openSUSE, sponsored by the NLnet NGI0 fund.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ else
 	DEB_PACKAGE_NAME := $(BINARY_NAME)
 endif
 
-DATE          := $(shell date -u '+%Y-%m-%d-%H%M UTC')
+DATE          := $(shell date -u -r RELEASE_NOTES '+%Y-%m-%d-%H%M UTC')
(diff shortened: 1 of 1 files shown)


DevinCarr:
I think for this repo, we update the RELEASE_NOTES file with every release, so if you are looking for a stable timestamp to reference for the build date, it would be better to use that file. How does that sound?
Otherwise, this change seems fine and I'll be happy to help merge this.

bmwiedemann:
updated the commit

DevinCarr (approved):

bmwiedemann:
So is something missing or is it good to merge?

PR #1292, opened 2024-07-13
Refactor tunneldns dnsserver configuration

### LINKED ISSUE #745: proxy-dns is [possibly] caching negative responses
Using certbot to renew certificates using dns-01 challenge. The machine that’s running certbot client is using cloudflared as a dns-proxy for DoH.

In the process of renewing a certificate, certbot inserts a [TXT](https://community.cloudflare.com/t/adding-dns-records/52718) record _acme-challenge.host.domain using the Cloudflare API. This record is correctly created.

Certbot then proceeds to poll for this value to ensure it’s queryable before asking the certificate provider to generate a certificate. The first query for this record results in an NXDOMAIN result since the record has not yet propagated to queryable DNS servers.

Subsequent polls, are getting a cached negative result. You can see the TTL decreasing on manual queries. The initial TTL on the negative result appears to be 30 minutes.
Certbot times-out after 5 minutes of polling for the record, so certificate renewal never succeeds. Restarting the cloudflared process clears the cache and later queries successfully get the correct [TXT](https://community.cloudflare.com/t/adding-dns-records/52718) record result.

I don’t know whether this is a change to cloudflared or whether upstream authoritative servers are now sending SOA TTLs for negative responses resulting in cloudflared caching the negative result. The effect is that using certbot with dns-01 challenge, via Cloudflare + cloudflared proxy-dns does not work.

This machine had been renewing certificates happily. The most recent renewal failed and the certificate expired. Suggesting that something changed within the last 3 months. The version of cloudflared being used is from Aug 22.

Note: I’m actually using the built-in certificate issuance from Caddy, but used certbot here as a more familiar term.
Also posted here: https://community.cloudflare.com/t/cloudflared-proxy-dns-is-caching-negative-responses/414454

alex4108:
Refactored the tunneldns package's dnsserver listener configuration.  This provides a path to disabling response caching as needed to fix #745

### DIFF
--- a/tunneldns/tunnel.go
+++ b/tunneldns/tunnel.go
@@ -94,14 +94,15 @@ func CreateListener(address string, port uint16, upstreams []string, bootstraps 
 	
 	
 	// Create a local cache with HTTPS proxy plugin
-	chain := cache.New()
-	chain.Next = ProxyPlugin{
+	proxyPlugin := ProxyPlugin{
 		Upstreams: upstreamList,
-	}	
+	}
+	chain := cache.New()
+	chain.Next = proxyPlugin	
 
 	// Optionally disable http response caching
 	if os.Getenv("DISABLE_TUNNELDNS_CACHE") == "true" { 
-		chain.Next = ProxyPlugin{}
+		chain = proxyPlugin 
 	}
 	
 	// Format an endpoint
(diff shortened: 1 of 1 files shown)


chungthuang:
Hi @alex4108, thank you for making this refactoring. We are in the process of deciding what will be the future of the DNS feature in cloudflared, so all developments are currently on hold.

Gu-myhome:
繼續玩 我家網路早已經被高層單位注意了，反正我沒差 我註冊是小朋友的名字 你們已經觸犯你們歐美法條 照片我都存著 你們繼續

chungthuang ***@***.***>於 2024年8月6日 週二，23:12寫道：

PR #1316, opened 2024-08-25
Adds a new command line for tunnel run for token file

### LINKED ISSUE #645: Allow passing tunnel token via the file system.
**Describe the feature you'd like**
I currently use named tunnels and run in docker containers. I would like to be able to use [docker secrets](https://docs.docker.com/engine/swarm/secrets/) with the containers. This is currently not possible because docker secrets are passed as a file and there is no way to read the file. It would be nice to be able to have `--secret-file` option, i.e. `cloudflared tunnel run --token --secret-file` or `cloudflared tunnel run --token </path/to/token/file>`.

**Describe alternatives you've considered**
Typically, you use `cat` in containers to read the files, but the rootless container does not have the cat executable so unable to use secrets that way

**Additional context**
Similar to #232. 

I would be happy to work on this if it is something that is worth implementing.

Cyb3r-Jak3:
Adds a new command line flag for `tunnel run` which allows a file to be read for the token. I've left the token command line argument with priority. 

Would resolve #645

### DIFF
--- a/cmd/cloudflared/tunnel/subcommands.go
+++ b/cmd/cloudflared/tunnel/subcommands.go
@@ -35,6 +35,7 @@ const (
 	CredFileFlag         = "credentials-file"
 	CredContentsFlag     = "credentials-contents"
 	TunnelTokenFlag      = "token"
+	TunnelTokenFileFlag  = "token-file"
 	overwriteDNSFlagName = "overwrite-dns"
 
 	LogFieldTunnelID = "tunnelID"
@@ -113,8 +114,13 @@ var (
 	})
 	tunnelTokenFlag = altsrc.NewStringFlag(&cli.StringFlag{
 		Name:    TunnelTokenFlag,
-		Usage:   "The Tunnel token. When provided along with credentials, this will take precedence.",
+		Usage:   "The Tunnel token. When provided along with credentials, this will take precedence. Also takes precedence over token-file",
 		EnvVars: []string{"TUNNEL_TOKEN"},
+	})
+	tunnelTokenFileFlag = altsrc.NewStringFlag(&cli.StringFlag{
+		Name:    TunnelTokenFileFlag,
+		Usage:   "Filepath at which to read the tunnel token. When provided along with credentials, this will take precedence.",
+		EnvVars: []string{"TUNNEL_TOKEN_FILE"},
 	})
 	forceDeleteFlag = &cli.BoolFlag{
 		Name:    "force",
@@ -613,6 +619,7 @@ func buildRunCommand() *cli.Command {
 		selectProtocolFlag,
 		featuresFlag,
 		tunnelTokenFlag,
+		tunnelTokenFileFlag,
 		icmpv4SrcFlag,
 		icmpv6SrcFlag,
 	}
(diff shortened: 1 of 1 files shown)


janani-cr:
@Cyb3r-Jak3 Can you please elaborate on your need to store the tunnel token in a local file? Any details of your workflow would be helpful.

Cyb3r-Jak3:
@janani-cr The goal would be to use [docker secrets](https://docs.docker.com/engine/swarm/secrets/) to pass the token to cloudflared. Docker secrets are passed as a file to the container. Standard practice for other images to have `_FILE` suffix so secrets can be used.

rakino:
This feature is also desired for running cloudflared on Guix System or NixOS, since service definitions are done at build time and written to the store which is world-readable.

Only specifying path to a keyfile outside the store can avoid the token being included in it.

masonfx:
This would be great to get merged. I've been looking for a more user-friendly (see: not a workaround) way to provide the tunnel token as a Docker secret.

KalenWessel:
Would love to see this released soon. All of my other services are able to use the `_FILE` convention with docker secrets. Cloudflare tunnel is the last service holding out on me.

david-plugge:
Yes please!

cmd/cloudflared/tunnel/subcommands.go
```diff
@@ -736,12 +743,25 @@ func runCommand(c *cli.Context) error {
...
+	// Check if tokenStr is blank before checking for tokenFile
+	if tokenStr == "" {
+		if tokenFile := c.String(TunnelTokenFileFlag); tokenFile != "" {
+			data, err := os.ReadFile(tokenFile)
+			if err != nil {
+				return cliutil.UsageError("Failed to read token file: " + err.Error())
+			}
+			tokenStr = strings.TrimSpace(string(data))
+			if _, err := ParseToken(tokenStr); err != nil {
+				return cliutil.UsageError("Failed to parse token file: " + err.Error())
```
lazysegtree, line 756: The user facing error message should be consistent in case of invalid token via token command line arg and via token file arg.
Right now, its `Failed to parse token file: <error>` for invalid token via file, and `Provided Tunnel token is not valid.` for invalid token via command line arg, which is inconsistent.

lazysegtree:
@Cyb3r-Jak3 I have another concern about this. `os.ReadFile`  reads the entire file into the memory, so should there be some sort of check on file size before reading the entire file ?
We could use `os.Stat` to get FileInfo object, and `FileInfo.Size()` to get file size, and apply a size check ( maybe limit files to < 10 MB )

- https://pkg.go.dev/os#File.Stat
- https://pkg.go.dev/io/fs#FileInfo


Not sure if this is needed or not.

Cyb3r-Jak3:
> @Cyb3r-Jak3 I have another concern about this. os.ReadFile reads the entire file into the memory, so should there be some sort of check on file size before reading the entire file ? 

I thought about this, but the config isn't checked for size, I didn't add it here.

cmd/cloudflared/tunnel/subcommands.go
lazysegtree: @Cyb3r-Jak3 Could you make the error message consistent ?

Domini:
Please, give this issue some love.

lazysegtree:
From the open PRs and no response from maintainers, I dont know if the current maintainers of the repo even care about contributions open source from open source community. God knows what are they thinking.

jcsf (approved):

PR #1335, opened 2024-10-08
Fix 'service install' on Mac OS

povilasb:
`cloudflared service install` is not working as documented at https://github.com/cloudflare/cloudflared/issues/327

When **cloudflared** is started with [Mac OS launch agent](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPSystemStartup/Chapters/CreatingLaunchdJobs.html) it errors with:
```
Use `cloudflared tunnel run` to start tunnel {UUID}
Use `cloudflared tunnel run` to start tunnel {UUID}
Use `cloudflared tunnel run` to start tunnel {UUID}
```

A fix is suggested at https://github.com/cloudflare/cloudflared/issues/327#issuecomment-1763670368
And acknowledged by https://community.cloudflare.com/t/tunnel-error-on-macos-when-installed-as-service/339543/4

However, collective costs keep compounding.

The fix seems simple at least for my use case where only a single tunnel is used. But Not sure if there's other negative side effects.
`cloudflared service install` produces such `.plist`
```xml
cat ~/Library/LaunchAgents/com.cloudflare.cloudflared.plist
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
        <dict>
                <key>Label</key>
                <string>com.cloudflare.cloudflared</string>
                <key>ProgramArguments</key>
                <array>
                        <string>cloudflared</string>
                        <string>tunnel</string>
                        <string>run</string>
                </array>
                <key>RunAtLoad</key>
                <true/>
                <key>StandardOutPath</key>
                <string>/Users/povilas/Library/Logs/com.cloudflare.cloudflared.out.log</string>
                <key>StandardErrorPath</key>
                <string>/Users/povilas/Library/Logs/com.cloudflare.cloudflared.err.log</string>
                <key>KeepAlive</key>
                <dict>
                        <key>SuccessfulExit</key>
                        <false/>
                </dict>
                <key>ThrottleInterval</key>
                <integer>5</integer>
        </dict>
</plist>
```

as opposed to
```xml
                <array>
                        <string>cloudflared</string>
                </array>
```

### DIFF
--- a/cmd/cloudflared/common_service.go
+++ b/cmd/cloudflared/common_service.go
@@ -24,7 +24,8 @@ func getServiceExtraArgsFromCliArgs(c *cli.Context, log *zerolog.Logger) ([]stri
 		// currently, we only support extra args for token
 		return buildArgsForToken(c, log)
 	} else {
-		// empty extra args
-		return make([]string, 0), nil
+		return []string{
+			"tunnel", "run",
+		}, nil
 	}
 }


cmd/cloudflared/common_service.go
```diff
@@ -24,7 +24,8 @@ func getServiceExtraArgsFromCliArgs(c *cli.Context, log *zerolog.Logger) ([]stri
 		// currently, we only support extra args for token
 		return buildArgsForToken(c, log)
 	} else {
-		// empty extra args
-		return make([]string, 0), nil
+		return []string{
```
DevinCarr, line 27: Instead of limiting this to only be able to run with tunnel parameters, could we improve this to allow any tunnel sub-commands to be passed in?
For example, if someone wanted to run `cloudflared access tcp` as a service, we should allow them to also setup using this command similar to the following:
```
cloudflared service install -- access tcp --hostname test.example.com --url localhost:8080
```

We would still like to allow a user to provide just their tunnel token, so we can still leave and try and parse that first to insert the `tunnel run --token <TOKEN>` command, but if that fails, we should just provide whatever the user provides as a subcommand to cloudflared.

I think that solves the problem you are trying to solve here while also expanding the functionality to be expressive for other folks that might try something similar. What do you think?

balaclava-guy:
Where did this end up going? I am just trying to get cloudflared to persist my tunnel after reboots, cant believe how painful this is

dfang:
https://github.com/cloudflare/cloudflared/issues/327

merge this pr pleasssssssssse or do sth else to fix this

Issue #1356, opened 2024-11-11
🐛 Missing verison number in RPM/DEB package filenames


Mikrz:
**Describe the bug**
The missing version number in the package filename causes issues in package repository mirrors such as Pulp3 as they are unable to maintain a version history. Additionally it breaks http caching proxies such as Squid that rely on the filename to determine cache availability, causing new package versions to be considered unchanged in the cache, leading to package installation failure due to a mismatch between fingerprint and content, rather than a simple cache miss.

**Expected behavior**
Builds should include version numbers in RPM and DEB packages.

PR #1357, opened 2024-11-11
Add package versions to the rpm and deb filenames

Mikrz:
Package mirrors like Pulp3 and caching proxies like Squid do not like rpm/deb filenames that are lacking a version number. This adds said version number, less the commit hash, to allow these tools to be happy.

Issue: #1356


chungthuang:
Hi @Mikrz, thank you for this suggestion. However, we will need to test this internally with our release pipelines, but we can't prioritize this right now. We will add this to the backlog.

lexfrei:
@chungthuang any news?

Issue #1294, opened 2024-07-15
🐛 Can't use Unix Sockets in Quick Tunnel mode


radiumatic:
**Describe the bug**

When you try to use a unix socket (`--unix-socket`) without a domain configured, `cloudflared` still asks for the --url parameter.

**To Reproduce**
1. Get a program to run on a unix socket
2. Run this comand
```bash
cloudflared tunnel --unix-socket /var/run/X.socket```
```
If it's an issue with Cloudflare Tunnel:
4. Tunnel ID : 
5. cloudflared config: 

**Expected behavior**

`cloudflared` tunneling the unix socket.

**Environment and versions**
 - OS: Debian
 - Architecture: X64
  - Version: 

**Logs and errors**
If applicable, add logs or errors to help explain your problem.

```
$ > ./cloudflared-linux-amd64 tunnel --unix-socket /var/run/X.socket               
You did not specify any valid additional argument to the cloudflared tunnel command.                                                                                                                           
If you are trying to run a Quick Tunnel then you need to explicitly pass the --url flag.
Eg. cloudflared tunnel --url localhost:8080/.

Please note that Quick Tunnels are meant to be ephemeral and should only be used
 for testing purposes.
For production usage, we recommend creating Named Tunnels. (https://developers.c
loudflare.com/cloudflare-one/connections/connect-apps/install-and-setup/tunnel-g
uide/)
```

**Additional context**

PR #1367, opened 2024-12-10
Fix: Can't use Unix Sockets in Quick Tunnel mode

rhyswilliamsza:
From https://github.com/cloudflare/cloudflared/issues/1294:

> When you try to use a unix socket (--unix-socket) without a domain configured, cloudflared still asks for the --url parameter.

This PR aims to address the above by adding `unix-socket` as an accepted flag for spawning quick tunnels. I unfortunately couldn't find an elegant way to mock out the `RunQuickTunnel` command, so I had to add a small work around. Please let me know if there is a better standard for this.

Thanks!

### DIFF
--- a/cmd/cloudflared/tunnel/cmd.go
+++ b/cmd/cloudflared/tunnel/cmd.go
@@ -215,6 +215,7 @@ var (
 		"overwrite-dns",
 		"help",
 	}
+	runQuickTunnel = RunQuickTunnel
 )
 
 func Flags() []cli.Flag {
@@ -313,9 +314,9 @@ func TunnelCommand(c *cli.Context) error {
 	// Run a quick tunnel
 	// A unauthenticated named tunnel hosted on <random>.<quick-tunnels-service>.com
 	// We don't support running proxy-dns and a quick tunnel at the same time as the same process
-	shouldRunQuickTunnel := c.IsSet("url") || c.IsSet(ingress.HelloWorldFlag)
+	shouldRunQuickTunnel := c.IsSet("url") || c.IsSet("unix-socket") || c.IsSet(ingress.HelloWorldFlag)
 	if !c.IsSet("proxy-dns") && c.String("quick-service") != "" && shouldRunQuickTunnel {
-		return RunQuickTunnel(sc)
+		return runQuickTunnel(sc)
 	}
 
 	// If user provides a config, check to see if they meant to use `tunnel run` instead
--- a/cmd/cloudflared/tunnel/cmd_test.go
+++ b/cmd/cloudflared/tunnel/cmd_test.go
@@ -21,40 +21,34 @@ func TestHostnameFromURI(t *testing.T) {
 
 func TestShouldRunQuickTunnel(t *testing.T) {
 	tests := []struct {
-		name              string
-		flags             map[string]string
-		expectQuickTunnel bool
-		expectError       bool
+		name        string
+		flags       map[string]string
+		expectError bool
 	}{
 		{
-			name:              "Quick tunnel with URL set",
-			flags:             map[string]string{"url": "http://127.0.0.1:8080", "quick-service": "https://fakeapi.trycloudflare.com"},
-			expectQuickTunnel: true,
-			expectError:       false,
+			name:        "Quick tunnel with URL set",
+			flags:       map[string]string{"url": "http://127.0.0.1:8080", "quick-service": "https://fakeapi.trycloudflare.com"},
+			expectError: false,
 		},
 		{
-			name:              "Quick tunnel with unix-socket set",
-			flags:             map[string]string{"unix-socket": "/tmp/socket", "quick-service": "https://fakeapi.trycloudflare.com"},
-			expectQuickTunnel: true,
-			expectError:       false,
+			name:        "Quick tunnel with unix-socket set",
+			flags:       map[string]string{"unix-socket": "/tmp/socket", "quick-service": "https://fakeapi.trycloudflare.com"},
+			expectError: false,
 		},
 		{
-			name:              "Quick tunnel with hello-world flag",
-			flags:             map[string]string{"hello-world": "true", "quick-service": "https://fakeapi.trycloudflare.com"},
-			expectQuickTunnel: true,
-			expectError:       false,
+			name:        "Quick tunnel with hello-world flag",
+			flags:       map[string]string{"hello-world": "true", "quick-service": "https://fakeapi.trycloudflare.com"},
+			expectError: false,
 		},
 		{
-			name:              "Quick tunnel with proxy-dns (invalid combo)",
-			flags:             map[string]string{"url": "http://127.0.0.1:9090", "proxy-dns": "true", "quick-service": "https://fakeapi.trycloudflare.com"},
-			expectQuickTunnel: false,
-			expectError:       true,
+			name:        "Quick tunnel with proxy-dns (invalid combo)",
+			flags:       map[string]string{"url": "http://127.0.0.1:9090", "proxy-dns": "true", "quick-service": "https://fakeapi.trycloudflare.com"},
+			expectError: true,
 		},
 		{
-			name:              "No quick-service set",
-			flags:             map[string]string{"url": "http://127.0.0.1:9090"},
-			expectQuickTunnel: false,
-			expectError:       true,
+			name:        "No quick-service set",
+			flags:       map[string]string{"url": "http://127.0.0.1:9090"},
+			expectError: true,
 		},
 	}
 
@@ -79,11 +73,10 @@ func TestShouldRunQuickTunnel(t *testing.T) {
 
 			// Validate
 			if tt.expectError {
+				assert.False(t, mockCalled)
 				require.Error(t, err)
-			} else if tt.expectQuickTunnel {
-				assert.True(t, mockCalled)
-				require.NoError(t, err)
 			} else {
+				assert.True(t, mockCalled)
 				require.NoError(t, err)
 			}
 		})
--- a/cmd/cloudflared/tunnel/cmd_test.go
+++ b/cmd/cloudflared/tunnel/cmd_test.go
@@ -1,9 +1,12 @@
 package tunnel
 
 import (
+	"flag"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/urfave/cli/v2"
 )
 
 func TestHostnameFromURI(t *testing.T) {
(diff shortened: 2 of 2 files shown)


cmd/cloudflared/tunnel/cmd.go
```diff
@@ -313,9 +314,9 @@ func TunnelCommand(c *cli.Context) error {
 	// Run a quick tunnel
 	// A unauthenticated named tunnel hosted on <random>.<quick-tunnels-service>.com
 	// We don't support running proxy-dns and a quick tunnel at the same time as the same process
-	shouldRunQuickTunnel := c.IsSet("url") || c.IsSet(ingress.HelloWorldFlag)
+	shouldRunQuickTunnel := c.IsSet("url") || c.IsSet("unix-socket") || c.IsSet(ingress.HelloWorldFlag)
 	if !c.IsSet("proxy-dns") && c.String("quick-service") != "" && shouldRunQuickTunnel {
-		return RunQuickTunnel(sc)
+		return runQuickTunnel(sc)
```
rhyswilliamsza, line 319: I was hoping to avoid changes here as far as possible, but I'm happy to refactor this into a separate function if this is too hacky.

cmd/cloudflared/tunnel/cmd_test.go
```diff
@@ -15,3 +18,76 @@ func TestHostnameFromURI(t *testing.T) {
 	assert.Equal(t, "", hostnameFromURI("trash"))
 	assert.Equal(t, "", hostnameFromURI("https://awesomesauce.com"))
 }
+
+func TestShouldRunQuickTunnel(t *testing.T) {
```
rhyswilliamsza, line 22: I've scoped this specifically to quick tunnels, so it won't test falling through to other tunnel types.

cmd/cloudflared/tunnel/cmd.go
```diff
@@ -313,9 +314,9 @@ func TunnelCommand(c *cli.Context) error {
 	// Run a quick tunnel
 	// A unauthenticated named tunnel hosted on <random>.<quick-tunnels-service>.com
 	// We don't support running proxy-dns and a quick tunnel at the same time as the same process
-	shouldRunQuickTunnel := c.IsSet("url") || c.IsSet(ingress.HelloWorldFlag)
+	shouldRunQuickTunnel := c.IsSet("url") || c.IsSet("unix-socket") || c.IsSet(ingress.HelloWorldFlag)
```
rhyswilliamsza, line 317: I haven't done any side-effect analysis to check whether anyone could be relying on `--unix-socket` falling through to `tunnel run`, but I believe it should be okay.

cmd/cloudflared/tunnel/cmd.go
lazysegtree: What is the point of this change ? 
`RunQuickTunnel` function is basically just aliased to `runQuickTunnel`.

lazysegtree: @rhyswilliamsza 
What is the point of this change ?
You have just aliased `RunQuickTunnel` to `runQuickTunnel`. Why ?

rhyswilliamsza: Hi @lazysegtree !

Thanks for taking a look. I've unfortunately had to do this to allow injection of mock `RunQuickTunnel`'s during testing (see cmd_test#61). Please let me know if there is a different way for me to be doing this!

cmd/cloudflared/tunnel/cmd.go
lazysegtree: I tried going through the code. 
In `TunnelCommand` function in cmd.go, if `--name` is not specified, only two non failure scenarios exists - quick tunnel or proxy dns. 
And in case of proxy-dns, I did not found a possible use of `--unix-socket` anywhere, as much I could understand the code flow.

So, looks good.

cmd/cloudflared/tunnel/cmd.go
lazysegtree: @rhyswilliamsza Got it. The test cases still seems pretty hacky (assigning `runQuickTunnel` to a new mock function, and defering the call to reset it). I would try to figure out if there is a cleaner way to mock `RunQuickTunnel` with the `testify` package (We are already using it in our test cases). 

A way that is not hacky, but neither does it introduces too many changes in source.

Still yours might be the only way that introduces minimum changes in the source.

rhyswilliamsza: Yip, I agree but unfortunately couldn't find a better way to mock out using the existing testing framework... I'll take another look after work today.

lazysegtree: @rhyswilliamsza We could make `TunnelCommand` function a wrapper, that calls the actual function with a `QuickTunnerRunner` parameter. Then we could use a diff mock implementation in test.

Thats a bit more intrusive on the source code, but less hacky. 

Tell me what you think about this.

<details>
<summary>Testcase run and diff</summary>

```
➜  ~/Workspace/Other_proj/rhyswilliamsza/cloudflared/cmd/cloudflared/tunnel git:(fix/unix-socket-quick-tunnel) ✗ [9:12:02] cfgo test -v -run "TestShouldRunQuickTunnel"
=== RUN   TestShouldRunQuickTunnel
=== RUN   TestShouldRunQuickTunnel/Quick_tunnel_with_URL_set
=== RUN   TestShouldRunQuickTunnel/Quick_tunnel_with_unix-socket_set
=== RUN   TestShouldRunQuickTunnel/Quick_tunnel_with_hello-world_flag
=== RUN   TestShouldRunQuickTunnel/Quick_tunnel_with_proxy-dns_(invalid_combo)
=== RUN   TestShouldRunQuickTunnel/No_quick-service_set
--- PASS: TestShouldRunQuickTunnel (0.00s)
    --- PASS: TestShouldRunQuickTunnel/Quick_tunnel_with_URL_set (0.00s)
    --- PASS: TestShouldRunQuickTunnel/Quick_tunnel_with_unix-socket_set (0.00s)
    --- PASS: TestShouldRunQuickTunnel/Quick_tunnel_with_hello-world_flag (0.00s)
    --- PASS: TestShouldRunQuickTunnel/Quick_tunnel_with_proxy-dns_(invalid_combo) (0.00s)
    --- PASS: TestShouldRunQuickTunnel/No_quick-service_set (0.00s)
PASS
ok  	github.com/cloudflare/cloudflared/cmd/cloudflared/tunnel	1.251s
```

diff
```
➜  ~/Workspace/Other_proj/rhyswilliamsza/cloudflared/cmd/cloudflared/tunnel git:(fix/unix-socket-quick-tunnel) ✗ [9:16:05] git --no-pager diff
diff --git a/cmd/cloudflared/tunnel/cmd.go b/cmd/cloudflared/tunnel/cmd.go
index 24ec3817..21aff7b0 100644
--- a/cmd/cloudflared/tunnel/cmd.go
+++ b/cmd/cloudflared/tunnel/cmd.go
@@ -215,7 +215,6 @@ var (
 		"overwrite-dns",
 		"help",
 	}
-	runQuickTunnel = RunQuickTunnel
 )

 func Flags() []cli.Flag {
@@ -287,7 +286,14 @@ See https://developers.cloudflare.com/cloudflare-one/connections/connect-apps/in
 	}
 }

+// This is so that we can mock QuickTunnelRunner for TunnelCommand test cases
+type QuickTunnelRunner func(*subcommandContext) error
+
 func TunnelCommand(c *cli.Context) error {
+	return TunnelCommandImpl(c, RunQuickTunnel)
+}
+
+func TunnelCommandImpl(c *cli.Context, quickTunnelRunner QuickTunnelRunner) error {
 	sc, err := newSubcommandContext(c)
 	if err != nil {
 		return err
@@ -316,7 +322,7 @@ func TunnelCommand(c *cli.Context) error {
 	// We don't support running proxy-dns and a quick tunnel at the same time as the same process
 	shouldRunQuickTunnel := c.IsSet("url") || c.IsSet("unix-socket") || c.IsSet(ingress.HelloWorldFlag)
 	if !c.IsSet("proxy-dns") && c.String("quick-service") != "" && shouldRunQuickTunnel {
-		return runQuickTunnel(sc)
+		return quickTunnelRunner(sc)
 	}

 	// If user provides a config, check to see if they meant to use `tunnel run` instead
diff --git a/cmd/cloudflared/tunnel/cmd_test.go b/cmd/cloudflared/tunnel/cmd_test.go
index faf1de00..0ac039a0 100644
--- a/cmd/cloudflared/tunnel/cmd_test.go
+++ b/cmd/cloudflared/tunnel/cmd_test.go
@@ -55,10 +55,8 @@ func TestShouldRunQuickTunnel(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			// Mock RunQuickTunnel Function
-			originalRunQuickTunnel := runQuickTunnel
-			defer func() { runQuickTunnel = originalRunQuickTunnel }()
 			mockCalled := false
-			runQuickTunnel = func(sc *subcommandContext) error {
+			runQuickTunnel := func(sc *subcommandContext) error {
 				mockCalled = true
 				return nil
 			}
@@ -69,7 +67,7 @@ func TestShouldRunQuickTunnel(t *testing.T) {
 			context := cli.NewContext(app, set, nil)

 			// Call TunnelCommand
-			err := TunnelCommand(context)
+			err := TunnelCommandImpl(context, runQuickTunnel)

 			// Validate
 			if tt.expectError {
➜  ~/Workspace/Other_proj/rhyswilliamsza/cloudflared/cmd/cloudflared/tunnel git:(fix/unix-socket-quick-tunnel) ✗ [9:17:36]
```

</details>

rhyswilliamsza: Thanks, @lazysegtree! Sorry for the delay, has been a busy week. 

I was hoping to change the source as little as possible, however I agree that it's better to do this in a more structured fashion. I like your approach better than the silly 'swapping in a function' method, even if it does require additional changes. Let me take a look at this again & I'll stage the change to this PR :)

lazysegtree: @rhyswilliamsza Thanks for adding me as collaborator. I have committed the changes.

jcsf:
Hi,
Thank you for your pull request and for your interest in improving the project. We appreciate the effort you put into adding support for Unix sockets in quick tunnels. However, quick tunnels are intended as a simplified, demo-oriented feature with limited scope. To keep them easy to use and maintain, we aim to support only a minimal and consistent set of options. For more advanced use cases like Unix sockets, we recommend using regular tunnels, which are designed to handle more complex configurations.

For this reason, we won’t be able to accept the PR, but we’re grateful for your contribution and hope you’ll continue to share ideas and improvements with us in the future.

Regards

PR #1371, opened 2024-12-14
fix: CVE-2024-45337: cloudflared: golang.org/x/crypto/ssh: Misuse of

hillwoodroc:
ServerConfig.PublicKeyCallback may cause authorization bypass in golang.org/x/crypto


jcsf:
Hi,
We have bumped cloudflared to use go1.24 and x/crypto v0.38 so this CVE is fixed.

Nonetheless, thank you for your contribution.

Issue #1389, opened 2025-01-06
🐛Illegal status code can be registered at `http_status`


kurochan:
**Describe the bug**
![2024-12-24-11 09 30@2x](https://github.com/user-attachments/assets/7008e25b-a625-436d-87e6-fc57564ca0df)

Weak validation of `http_status`, so incorrect can register http status.

**To Reproduce**
1. Configure `http_status:999`

**Expected behavior**
Only existing http status codes can be registered.

**Environment and versions**
 - OS: All
 - Architecture: All
 - Version: 2024.12.2

**Additional context**
After this, I plan to create a PR, but fixing this issue requires modifications not only to cloudflared but also to the Web UI( https://one.dash.cloudflare.com/ ).

lazysegtree:
Is there a possible security issue due to invalid HTTP status code ?

lazysegtree:
Also there are some "Unofficial codes" that might be a valid use case (See [wiki]( https://en.wikipedia.org/wiki/List_of_HTTP_status_codes) ), for which go's `StatusText` function ([source](https://github.com/golang/go/blob/master/src/net/http/status.go)) would not return any text. 
- For example HTTP 218, 509, 529, and many more

(Here is a screenshot from wiki page about the additional unofficial codes, used by cloudflare too)
<img width="667" alt="image" src="https://github.com/user-attachments/assets/a42aae1c-ce08-40c2-b02d-fc0b1b8ae3ae" />

See - https://github.com/cloudflare/cloudflared/pull/1390#issuecomment-2600866730

lazysegtree:
In my opinion, this should not be done. We should allow any unofficial HTTP code, on any other customer HTTP code a user wants. This doesn't seem to pose any security risk.

PR #1390, opened 2025-01-06
Improve `http_status` validation

kurochan:
Currently, the validation of the http status input is a numeric range, so an invalid status code can be entered. Change to check if the http status code exists or not.
related to https://github.com/cloudflare/cloudflared/issues/1389 .

2024-12-24-11 09 30@2x

Even if this validation were added, it is still possible to set it from the Web UI( https://one.dash.cloudflare.com/ ), so that needs to be modified as well.

Those that exist as http status but are not normally used can be registered, but I think this should be handled by the web UI. 
(Personally, I don't see a problem with using the joke status code because it's funny.)
image(1)

### DIFF
--- a/ingress/ingress_test.go
+++ b/ingress/ingress_test.go
@@ -247,10 +247,18 @@ ingress:
 			wantErr: true,
 		},
 		{
-			name: "Invalid HTTP status code",
+			name: "Invalid HTTP status code(8080)",
 			args: args{rawYAML: `
 ingress:
  - service: http_status:8080
+`},
+			wantErr: true,
+		},
+		{
+			name: "Invalid HTTP status code(666)",
+			args: args{rawYAML: `
+ingress:
+ - service: http_status:666
 `},
 			wantErr: true,
 		},
--- a/ingress/ingress.go
+++ b/ingress/ingress.go
@@ -3,6 +3,7 @@ package ingress
 import (
 	"fmt"
 	"net"
+	"net/http"
 	"net/url"
 	"regexp"
 	"strconv"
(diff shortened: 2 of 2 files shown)


lazysegtree:
@kurochan, this apparently creates a dependency about http status codes on the go source code.


- This creates a hard dependency on update of go codebase for any new HTTP status code among the unassigned http status code (https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml). Only the status code that are mentioned in https://github.com/golang/go/blob/master/src/net/http/status.go would stay valid. 

Also there are some "Unofficial codes" that might be a valid use case (See https://en.wikipedia.org/wiki/List_of_HTTP_status_codes), for which go's `StatusText` function would not return any text.
- For example HTTP 218, 509, 529, and many more

(Here is a screenshot from wiki page about the additional unofficial codes, used by cloudflare too)
image

ingress/ingress_test.go
```diff
@@ -247,10 +247,18 @@ ingress:
...
-			name: "Invalid HTTP status code",
+			name: "Invalid HTTP status code(8080)",
 			args: args{rawYAML: `
 ingress:
  - service: http_status:8080
+`},
+			wantErr: true,
+		},
+		{
+			name: "Invalid HTTP status code(666)",
```
vahan-sahakyan-op, line 258:
```suggestion
			name: "Invalid HTTP status code(777)",
```

DevinCarr:
https://github.com/cloudflare/cloudflared/issues/1389#issuecomment-2770304634

PR #1418, opened 2025-02-20
feat(Makefile): add support for arch mips

fumiama:
Usage:
```sh
TARGET_OS=linux GOARCH=mipsle TARGET_MIPS=softfloat make cloudflared
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -119,6 +119,11 @@ else
 	PACKAGE_ARCH := $(TARGET_ARCH)
 endif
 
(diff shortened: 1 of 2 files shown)


DevinCarr:
Thanks for this contribution, however, we do not currently suspect we will support packaging cloudflared for MIPS in the short term.

You should be able to support building cloudflared for MIPS without issue by updating your build command to:
```
TARGET_OS=linux GOARCH=mipsle GOMIPS=softfloat make cloudflared
```

PR #1446, opened 2025-04-08
Use numeric user in Dockerfiles

EpicStep:
Hi, this improvement allows 'cloudflared' to run in Kubernetes with 'runAsNonRoot: true' without requiring force change user ID to be numeric.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 # use a builder image for building cloudflare
 ARG TARGET_GOOS
 ARG TARGET_GOARCH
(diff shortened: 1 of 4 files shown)


jcsf:
Hi,
We have detected this issue and fixed it on the last release. The correct user id to be used is the one from the distroless images https://github.com/GoogleContainerTools/distroless/blob/main/common/variables.bzl#L18

Regardless, thank you for bringing awareness and contributing for this project. We will close this PR since it was using wrong user id's, but we appreciate your contribution.

Regards

PR #1453, opened 2025-04-16
fix go vet  non-constant format string in calls

crrodriguez:
multiple non-constant format string in calls fixed.


jcsf:
Hi,
Thank you for your pull request and for taking the time to contribute to the project.
We have linters in place to fix this type of issues, but we are doing it slowly and only on files that are changed, not on the ones from the past. This said, this will eventually get fixed by our normal process of development, so we won't be accepting this PR.

Regardless, thanks again for your contribution and interest in the project!

PR #1464, opened 2025-05-07
Fix: Adding support for namespace when collecting logs from Kubernetes deployments

nirav-chotai:
**Problem:** If tunnel is deployed in other namespace than `default` then `cloudflared tunnel diag` will fail to collect logs as there is no option to pass the `namespace` as an argument. 

**Solution:** If tunnel is deployed in K8s other than default namespace, then this will allow to specify namespace option to collect logs from when running `cloudflared tunnel diag`


jcsf:
Hi @nirav-chotai can you please rebase the PR and add some unit-tests, ensure that everything is compiling and passing all tests.

Thanks for your contribution

PR #1476, opened 2025-05-19
Add build UUID

### LINKED ISSUE #1460: 🐛 Local network permissions when running as non-root on macOS
**Describe the bug**

When I run `cloudflared` as a non-root user, installed with `cloudflared service install`, it can't access any backend service in the local network. Each access results in a log entry like this:

```
Unable to reach the origin service. The service may be down or it may not be responding to traffic from cloudflared: dial tcp 192.168.1.1:443: connect: no route to host
```

Any backend on the localhost is accessible, so the tunnel is working overall.

There are no network issues and I can reach this IP just from the host. After `launchctl unload ~/Library/LaunchAgents/com.cloudflare.cloudflared.plist` and `sudo cloudflared service install`, the same cloudflared binary with the same command line arguments is able to serve the backends on the local network.

**To Reproduce**
1. Configure `cloudflared` to run as a local user
2. Run `launchctl bootstrap gui/501 ~/Library/LaunchAgents/com.cloudflare.cloudflared.plist` to start it
3. Configure a backend on a different machine on the local network
4. Attempt to access the backend

**Expected behavior**

I expect `cloudflared` to behave identically when run as `root` or a different user.

**Environment and versions**
 - OS: macOS 15.4.1
 - Architecture: Apple Silicon
 - Version: 2025.04.2

**Logs and errors**
See above

**Additional context**
When debugging the issue over Apple Remote Desktop, I got a popup asking whether `cloudflared` should be allowed local network access. I clicked "yes" and verified that `/opt/homebrew/Cellar/cloudflared/2025.4.0/bin/cldouflared` is enabled in System Preferences / Privacy & Security / Local Network.

I assume this is related to macOS's security mechanisms.

IngmarStein:
Recent macOS versions require user permissions to access the local network. This can only be granted for binaries which have a LC_UUID load command. Go 1.24 adds the UUID by default, but the fix was backported to Go 1.22.9 and 1.23 where it requires passing `-B gobuildid` to the linker.

cloudflared uses Cloudflare's Go fork which is based on Go 1.22.5 and doesn't have the backport. Because of this, cloudflared can't talk to any machine on the local network (no route to host) when it's runnning as a non-root user.

This workaround updates the Go fork to go1.22.12-devel-cf and adds `-B gobuildid` so that the binary now has a UUID:

```
❯ dwarfdump --uuid /opt/homebrew/bin/cloudflared
UUID: F2088E17-03A1-34B1-F12C-0DF01F00BA32 (arm64) /opt/homebrew/bin/cloudflared
```

Fixes #1460

### DIFF
--- a/.teamcity/install-cloudflare-go.sh
+++ b/.teamcity/install-cloudflare-go.sh
@@ -3,6 +3,6 @@
 cd /tmp
 git clone -q https://github.com/cloudflare/go
 cd go/src
-# https://github.com/cloudflare/go/tree/af19da5605ca11f85776ef7af3384a02a315a52b is version go1.22.5-devel-cf
-git checkout -q af19da5605ca11f85776ef7af3384a02a315a52b
+# https://github.com/cloudflare/go/tree/37bc41c6ff79507200a315b72834fce6ca427a7e is version go1.22.12-devel-cf
+git checkout -q 37bc41c6ff79507200a315b72834fce6ca427a7e
 ./make.bash
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,7 @@ ifeq ($(FIPS), true)
 	VERSION_FLAGS := $(VERSION_FLAGS) -X "main.BuildType=FIPS"
 endif
 
-LDFLAGS := -ldflags='$(VERSION_FLAGS) $(LINK_FLAGS)'
+LDFLAGS := -ldflags='$(VERSION_FLAGS) $(LINK_FLAGS) -B gobuildid'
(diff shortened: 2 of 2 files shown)


IngmarStein:
CC @jcsf

IngmarStein:
@jcsf Even better, thanks!

Not sure if you're aware, but homebrew-core started using the system-provided Go rather than your fork: https://github.com/Homebrew/homebrew-core/pull/223905
You might want to reach out in case this leads to a fragmented user experience.

PR #1483, opened 2025-05-29
A couple of grammar fixes

gameroman:


jcsf:
Hi,
Thank you for your pull request and for taking the time to contribute to the project.
This PR changes vendoring files which are not supposed to be changed and will be removed in the future. Since this is simply fixing a typo, we will not be accepting it since we can fix it while doing other changes on the code.

Regardless, thanks again for your contribution and interest in the project!

PR #1494, opened 2025-06-22
handle http2 GOAWAY

ganey:
I've been experiencing the GOAWAY error, it comes in spikes and affects thousands of our users.

We run roughly 9 cloudflared, with 4gb ram, 4cpu each, lowest traffic is around 60 requests/s per cloudflared. We find these GOAWAY errors in the logs when we noticed a huge spike in 524 errors to our users.

Builds on top of this PR https://github.com/cloudflare/cloudflared/pull/1274

> Fixes https://github.com/cloudflare/cloudflared/issues/1273
> 
> Allows http2 requests to be retried only when a GOAWAY frame is received.
> 
> GOAWAY spec: https://datatracker.ietf.org/doc/html/rfc7540#section-6.8
> How GetBody is handled: https://github.com/golang/net/blob/master/http2/transport.go#L664


diogotorres97:
ping @ganey

ganey:
> ping @ganey 

hey, what do you need from me?

szmarczak:
I haven't looked into this deeply but AFAIK goaway with code 0 means it should shut down after all requests have ended (and responses fully received). Only error codes shut down the socket immediately.

PR #1510, opened 2025-07-29
Added --no-configupdate flag to reject remote config updates

felix-engelmann:
Running `cloudflared` allows anyone with access to the zero trust dashboard to probe services from the viewpoint of this host by adding malicious service urls. A simple way to prevent this security issue is to only allow locally defined `ingress` rules loaded from a config file.
To allow easy configuration via the web interface, the flag defaults to off and accepts updates pushed from the edge.
The change should be transparent to the control plane as the "correct" config version is returned as if it was applied.

There is no related issue with this PR, as I need this feature locally anyways.


nikitacano:
Appreciate the effort you've put into this, but we're hesitant to introduce a new configuration flag that would put a remotely-managed tunnel into a state where it ignores certain updates pushed from our control plane. This could lead to a confusing user experience and make troubleshooting very difficult.

We want to make sure we understand your specific security needs. Can you please elaborate on your use case? What is the specific threat model you are trying to protect against?

In the meantime, we want to make sure you're aware of [locally managed tunnels](https://developers.cloudflare.com/cloudflare-one/connections/connect-networks/do-more-with-tunnels/local-management/). If your goal is to completely prevent any remote management of your tunnel, you can run cloudflared in a local-only configuration. This is the most secure way to ensure that the tunnel's ingress rules and other configurations are only ever sourced from a local file. This mode is explicitly designed for this exact purpose.

felix-engelmann:
I understand and already notice the problem of inconsistent state between the control plane and the actual state. The service url in the dashboard is totally ignored.

The threat model I try to protect against is for my locally running cloudflared to receive a RPC update from edge servers to change the configuration. In the example of the http2 connection, I would like to disable this handler: https://github.com/cloudflare/cloudflared/blob/1cedefa1c22a74dd6371d4920398d553d01f1391/connection/http2.go#L174-L186

As I wanted to have it independent of the transport protocol, I implemented the change in the common `orchestrator.UpdateConfig`.

Thanks for pointing me to the locally managed tunnels. However, they use the [same command](https://developers.cloudflare.com/cloudflare-one/connections/connect-networks/do-more-with-tunnels/local-management/create-local-tunnel/#6-run-the-tunnel) to start a tunnel as a remotely managed one:

     cloudflared tunnel --config /path/your-config-file.yml run <UUID or NAME>


As the configuration file has the same structure, I did not find the code which disables the RPC functionality for locally managed tunnels. 

To prevent inconsistent state of the control plane, one possible solution could be to expand the `UpdateConfigurationResponse` struct to return the current *static* configuration. https://github.com/cloudflare/cloudflared/blob/1cedefa1c22a74dd6371d4920398d553d01f1391/tunnelrpc/pogs/configuration_manager.go#L79-L82
However this needs changes in the control plane to properly handle these replies and maybe disable the input fields in the dashboard.

nikitacano:
> Thanks for pointing me to the locally managed tunnels. However, they use the [same command](https://developers.cloudflare.com/cloudflare-one/connections/connect-networks/do-more-with-tunnels/local-management/create-local-tunnel/#6-run-the-tunnel) to start a tunnel as a remotely managed one:

The fundamental distinction between a locally managed tunnel and a remotely managed one is not just how they are started, but how they are created:
- Remotely managed tunnels are created via the Zero Trust dashboard or API. They are designed to be centrally controlled.
- Locally managed tunnels are created on the command line using `cloudflared tunnel create <NAME>`. This command generates a local configuration file and a certificate that are tied to that specific machine.

If a tunnel was created as locally-managed, it won't receive any configuration updates from our control plane as long as it is not migrated to remotely-managed:
Screenshot 2025-08-08 at 13 34 47

The only way a locally-managed tunnel can start receiving remote updates is if an administrator with sufficient privileges explicitly migrates it in the Zero Trust dashboard. This is a deliberate, manual action that requires admin-level access.

Given this, the most effective way to address your threat model — preventing unauthorized remote configuration changes — is to:
1. Use a locally managed tunnel. This is the intended and supported way to run a cloudflared instance that is not subject to remote control.
2. Secure your Zero Trust dashboard. As you noted, the ultimate vulnerability is an administrator with sufficient privileges. The best practice here is to enforce strong authentication policies on your Cloudflare account, such as requiring 2FA for all administrative logins. This provides a much more robust security layer than trying to introduce a "zombie" state in the cloudflared client itself.

Because a locally managed tunnel already provides the exact behaviour you are seeking without introducing the complexity of a new flag, we won't be moving forward with this PR. The risk of creating a confusing and difficult-to-support "zombie" state for remotely-managed tunnels is too high.

We highly recommend you create a new locally-managed tunnel for your use case. Thank you again for your contribution and for helping us make cloudflared more secure.

felix-engelmann:
Thank you for the detailed explanation of the difference between locally and centrally managed tunnels. I also fully understand that the increase in complexity is not worth maintaining for cloudflare.

The fact that a locally managed tunnel can be *remotely* changed into a centrally managed one already proves my point.
I have to trust the edge servers to behave honestly and only send RPC packets that a dashboard user initiated, which I do not. I only accept the edge servers' ability to inspect plain HTTP traffic after TLS termination for a locally restricted set of services. 

Going forward, I will try to maintain my fork.

PR #1512, opened 2025-08-02
fix s3 streaming uploads broken by go/net bump

### LINKED ISSUE #1511: Content-Length: 0 (NoBody) override breaks AWS S3 streaming uploads
**Describe the bug**
2025.7.0 causes AWS S3 (minio) chunked uploads to fail with `<Code>IncompleteBody</Code><Message>You did not provide the number of bytes specified by the Content-Length HTTP header.</Message>` due to incorrectly set `content-length: 0` by cloudflared.

**To Reproduce**
Upload a file with aws like `aws --endpoint-url=https://cloudflare-proxied-s3.example.com s3 cp file.png s3://bucket/file.png`

**Expected behavior**
Cloudflared should not set content-length to 0 for chunked transfers.

**Environment and versions**
 - OS: Ubuntu Linux 24.04.2 LTS
 - Architecture: AMD64
 - Version: 2025.7.0

**Logs and errors**
Below is an example request.
```
PUT /xxxx-docs/file.txt HTTP/1.1
Host: s3.xxxx.net
User-Agent: aws-cli/1.30.6 md/Botocore#1.37.28 ua/2.1 os/linux#6.8.0-64-generic md/arch#x86_64 lang/python#3.10.12 md/pyimpl#CPython m/N cfg/retry-mode#legacy botocore/1.37.28
Content-Length: 0
Accept-Encoding: gzip, br
Amz-Sdk-Invocation-Id: xxxx
Amz-Sdk-Request: attempt=1
Authorization: AWS4-HMAC-SHA256 Credential=xxxx/20250801/us-east-1/s3/aws4_request, SignedHeaders=content-encoding;content-type;host;x-amz-content-sha256;x-amz-date;x-amz-decoded-content-length;x-amz-sdk-checksum-algorithm;x-amz-trailer, Signature=xxxx
Cdn-Loop: cloudflare; loops=1
Cf-Connecting-Ip: xxxx
Cf-Ipcountry: XX
Cf-Ray: xxxx-CPH
Cf-Visitor: {"scheme":"https"}
Cf-Warp-Tag-Id: xxxx
Connection: keep-alive
Content-Encoding: aws-chunked
Content-Type: text/plain
X-Amz-Content-Sha256: STREAMING-UNSIGNED-PAYLOAD-TRAILER
X-Amz-Date: 20250801T234228Z
X-Amz-Decoded-Content-Length: 13
X-Amz-Sdk-Checksum-Algorithm: CRC32
X-Amz-Trailer: x-amz-checksum-crc32
X-Forwarded-For: xxxx
X-Forwarded-Proto: https
```

**Additional context**
I believe this is caused due to go net module upgrade. Uploads work fine on 2025.5.0.
New go net module sets content-length:0 if NoBody is set: https://github.com/golang/go/commit/450f3f608d409e2b3d76af071ec726efacbdd17b#diff-a50801014d4fc5373ebbddfedea688dbf90f9c68542b6d62b3ee4b2fa641e986R72

This is set by cloudflared in connection/quic_connection.go:370 , `(!isWebsocket && !isTransferEncodingChunked(req) && req.ContentLength == 0)`. When content-length is unset, it defaults to 0, and AWS uses `content-encoding: aws-chunked` instead of `transfer-encoding`, causing cloudflare to set NoBody, causing go net to explicitly set `content-length: 0`. Also checking for `content-encoding` seems to fix the issue. Will be creating a pull request. https://github.com/sh4dowb/cloudflared/commit/7705a8004b89dc1f889897aefc716d57c61f3a6d

sh4dowb:
fixes #1511 

go/net module was updated,
https://github.com/golang/go/commit/450f3f608d409e2b3d76af071ec726efacbdd17b#diff-a50801014d4fc5373ebbddfedea688dbf90f9c68542b6d62b3ee4b2fa641e986R72

cloudflared sets `NoBody` for non chunked and zero content length requests as seen below:
```
        // Go's client defaults to chunked encoding after a 200ms delay if the following cases are true:
        //   * the request body blocks
        //   * the content length is not set (or set to -1)
        //   * the method doesn't usually have a body (GET, HEAD, DELETE, ...)
        //   * there is no transfer-encoding=chunked already set.
        // So, if transfer cannot be chunked and content length is 0, we dont set a request body.
        if !isWebsocket && !isTransferEncodingChunked(req) && req.ContentLength == 0 {
                req.Body = http.NoBody
        }
```

However, when `content-length` is not set, it defaults to `0`, and AWS S3 uses `content-encoding: aws-chunked` instead of `transfer-encoding`, causing cloudflared to set `NoBody`.
With the go/net module 0.26.0->0.40.0 bump, it explicitly sets `content-length: 0` if request is NoBody, even when original request does not have a `content-length` header. This causes S3-compatible storages, at least minio, and potentially other applications to fail, because the uploaded body size does not match the header.

The fix simply searches the string `chunked` in `content-encoding` alongside `transfer-encoding`.

### DIFF
--- a/connection/quic_connection.go
+++ b/connection/quic_connection.go
@@ -386,10 +386,11 @@ func setContentLength(req *http.Request) error {
 }
 
 func isTransferEncodingChunked(req *http.Request) bool {
+	contentEncodingVal := req.Header.Get("Content-Encoding") // AWS S3 uses Content-Encoding: aws-chunked
 	transferEncodingVal := req.Header.Get("Transfer-Encoding")
 	// https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding suggests that this can be a comma
 	// separated value as well.
-	return strings.Contains(strings.ToLower(transferEncodingVal), "chunked")
+	return strings.Contains(strings.ToLower(transferEncodingVal), "chunked") || strings.Contains(strings.ToLower(contentEncodingVal), "chunked")
 }
 
 // A helper struct that guarantees a call to close only affects read side, but not write side.


jcsf (approved):

eliassoares:
Hey, I had the same issue, thanks @sh4dowb!

PR #1508, opened 2025-07-24
Patch and additional files for FreeBSD build

vrtareg:
This is intent to fix #1365 as per @NothingTooSerious suggestion in https://github.com/cloudflare/cloudflared/issues/1365#issuecomment-2678163738

Re-Creating pull to fix conflicts.

PR #1523, opened 2025-08-25
support for freebsd compile

swhe:
I tried to build it from freebsd and found some issues during build. Add the changes for two files and then pass the build in freebsd.


swhe:
close this as #1508 already done the same thing

PR #1527, opened 2025-09-06
docs: fix typo 'ot' -> 'to' in CHANGES.md

hexpreso:
Fixed typo in CHANGES.md on line 284 where 'ot' was corrected to 'to'.


jcsf:
Hi,

PR #1532, opened 2025-09-18
refactor: omit unnecessary reassignment

quantpoet:
The new version of Go has been optimized, and variables do not need to be reassigned.

For more info: https://tip.golang.org/wiki/LoopvarExperiment#does-this-mean-i-dont-have-to-write-x--x-in-my-loops-anymore


quantpoet:
@jcsf @chungthuang Hi, Could you please review this PR at your convenience? Thank you very much.

jcsf:
Hi,

Issue #1037, opened 2023-08-04
🐛 ERR Cannot determine default origin certificate path


firecow:
**Describe the bug**
When starting a `cloudflared tunnel run $TUNNEL_ID` command, I'm getting an ERR in the logs

```
ERR Cannot determine default origin certificate path. No file cert.pem in [~/.cloudflared ~/.cloudflare-warp ~/cloudflare-warp /etc/cloudflared /usr/local/etc/cloudflared]. You need to specify the origin certificate path by specifying the origincert option in the configuration file, or set TUNNEL_ORIGIN_CERT environment variable originCertPath=
```

**To Reproduce**
1. Just start a cloudflared tunnel run command

**Expected behavior**
No error log should be printed, about missing cert.pem file's, when it obviously isn't needed by the `cloudflared tunnel run` command.

joliveirinha:
Thanks for the report.

Would you like to open a PR to address this?

firecow:
> Thanks for the report.
> 
> Would you like to open a PR to address this? 

I can try 😁

aluferraz:
Same here, how did you solve it ?

firecow:
@aluferraz The issue is still open.

tranghaviet:
+ same here

djj106:
same issue

inRm3D:
a monkey patch

https://www.reddit.com/r/CloudFlare/comments/u4bwom/cloudflare_tunnels_error/

> Download the pem cert from Zero Trust for the argo tunnel. Upload it to /etc/cloudflared
>
> https://dash.cloudflare.com/argotunnel
>
> then select the zone and it should then download the cert

HyunSu1768:
It worked when I logged in with the root account, went to the directory where the cert file was located, and ran the command.

watford-ep:
What is the fix for this?

PR #1547, opened 2025-11-03
Prevent loading the origin certificate for remotely-managed tunnels

mocknen:
When running tunnels, an error log is output stating that the origin certificate cannot be found, even though it is not required for remotely managed tunnels.
This pull request modifies the behavior so that if a tunnel token is specified, the origin certificate will not be loaded.

Related to #1037.


jcsf:
Hi,
Thanks for your contribution. This has already been done by the team. So, we can close this.

Regards

PR #1564, opened 2025-12-05
Create generator-generic-ossf-slsa3-publish.yml

Pro2x9:


jcsf:
Hi,
This change has no information so we will just be closing it. If you want you can reopen in with the proper issue and description that contextualize the change.

PR #1574, opened 2025-12-21
Docker: use debian13 in the final image and update golang builder to 1.25.5

oittaa:
It would make sense to upgrade to the latest stable Debian image (13 / Trixie) in the final stage.


jcsf:
Hi,
Thank you for our contribution, but Go and Debian updates need to be done internally because we use our own go fork. Therefore I'm going to close this.

Regards

PR #1600, opened 2026-02-13
fix: simple typo

AfzalivE:
I saw the closed typo PRs so I didn't want to create this but it's been months and this typo keeps staring at me 😅.


jcsf:
Hi,

AfzalivE:
That's fair and I saw that comment on another PR too but seems like this typo had been missed.

PR #1601, opened 2026-02-17
feat: support multiple metrics addresses via repeated flag

IngmarStein:
## Summary
This PR updates the `--metrics` flag to accept multiple values, allowing the metrics server to bind to specific addresses (e.g., localhost IPv4 and IPv6) without exposing the endpoint to all interfaces via `[::]`.

## Changes
-   **CLI:** Changed `metrics` flag to `StringSliceFlag`. Users can now pass `--metrics` multiple times.
-   **Metrics Package:**
    -   `CreateMetricsListener` now accepts a slice of addresses and returns a slice of listeners.
    -   `ServeMetrics` now accepts a slice of listeners and serves them concurrently using a shared `http.Server`.
-   **Tunnel Command:**
    -   `StartServer` updated to initialize and close multiple listeners.
    -   `readyCommand` updated to check the first configured metrics address.
-   **Tests:** Added test cases for multiple listener creation.

## How to test
Run the tunnel with multiple metrics flags:
```bash
cloudflared tunnel run --metrics 127.0.0.1:2020 --metrics [::1]:2020 ...
```
Verify that the metrics endpoint is accessible on both addresses but not on external interfaces.

### DIFF
--- a/cmd/cloudflared/tunnel/cmd.go
+++ b/cmd/cloudflared/tunnel/cmd.go
@@ -445,13 +445,15 @@ func StartServer(
 		return err
 	}
 
-	metricsListener, err := metrics.CreateMetricsListener(&listeners, c.String("metrics"))
+	metricsListeners, err := metrics.CreateMetricsListener(&listeners, c.StringSlice("metrics"))
 	if err != nil {
 		log.Err(err).Msg("Error opening metrics server listener")
 		return errors.Wrap(err, "Error opening metrics server listener")
 	}
 
-	defer metricsListener.Close()
+	for _, l := range metricsListeners {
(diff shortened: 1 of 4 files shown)


jcsf:
Hi,
What is the problem you are trying to resolve here? Why is more than 1 metrics address needed? Can you give us an example or a more detailed explanation?

IngmarStein:
@jcsf if I use "--metrics 127.0.0.1:2020", cloudflared only listens on IPv4 (at least on macOS 26) and with "--metrics [::1]:2020", it listens on IPv6 only. I'd like to make the metrics endpoint available on both stacks, but without opening it up beyond localhost.

Issue #1334, opened 2024-10-01
🐛 The user running cloudflared process has a GID (group ID) that is not within ping_group_range


FezVrasta:
**Describe the bug**

I'm seeing the following error while running cloudflared through the official Docker image.

```
WRN The user running cloudflared process has a GID (group ID) that is not within ping_group_range. You might need to add that user to a group within that range, or instead update the range to encompass a group the user is already in by modifying /proc/sys/net/ipv4/ping_group_range. Otherwise cloudflared will not be able to ping this network error="Group ID 65532 is not between ping group 1 to 0"
```

After it, I also see:

```
WRN ICMP proxy feature is disabled error="cannot create ICMPv4 proxy: Group ID 65532 is not between ping group 1 to 0 nor ICMPv6 proxy: socket: permission denied"
failed to sufficiently increase receive buffer size (was: 208 kiB, wanted: 7168 kiB, got: 416 kiB). See https://github.com/quic-go/quic-go/wiki/UDP-Buffer-Sizes for details.
```

**To Reproduce**
1. Start cloudflared on a Kubernetes GKE cluster on Google Cloud

**Expected behavior**
These errors seem harmless, everything seem to be working fine, but I would like to understand why I'm seeing them and if I can do something to stop them.

**Environment and versions**
 - OS: [e.g. MacOS]
 - Architecture: [e.g. AMD, ARM]
 - Version: [e.g. 2022.02.0]

**Logs and errors**
If applicable, add logs or errors to help explain your problem.

**Additional context**

cyanidium:
I found this while searching for the same problem, but there is already a fix in Issue #1109 ([comment here](https://github.com/cloudflare/cloudflared/issues/1109#issuecomment-2234239403))

For the UDP buffer I'm running on baremetal and was able to update the sysctls as described in the [link from the logs](https://github.com/quic-go/quic-go/wiki/UDP-Buffer-Sizes).

FezVrasta:
I'm running on GKE and the cloudflared runs as a sidecar (initContainers), if I try to set the sysctls it tells me the syntax is invalid, I read it could because Autopilot doesn't support overriding sysctls? I also tried to update the GUID settings but it doesn't seem to make any difference?

```
apiVersion: apps/v1
kind: Deployment
metadata:
  annotations:
    kompose.cmd: kompose convert
    kompose.version: 1.33.0 (HEAD)
  labels:
    io.kompose.service: web-client
  name: web-client
spec:
[...]

PR #1602, opened 2026-02-23
fix(ci): enable ICMP tests on Linux and fix gotestfmt on macOS/Windows

mannie-exe:
Hey! I didn't see any issues for the CI/CD failures, so I'm not really sure if this is needed work, but I did get them working on my fork and wanted to share in case it's needed.

1. Linux: ICMP tests failure in GitHub Actions (#1334)
> ICMP tests fail because ping_group_range on GitHub Actions runners defaults to "1 0" (unprivileged ICMP disabled). Add sysctl step to allow unprivileged ICMP sockets.
2. macOS: gotestfmt panic (gotesttools/gotestfmt#64)
> go test -json emits linker warnings (LC_DYSYMTAB) as "Action":"build-output" JSON lines. gotestfmt v2.5.0 predates this action type (Go 1.21+) and panics in parseLine. Filter build-output lines before piping to gotestfmt.
3. Windows: /tmp/gotest.log path not found
> GO_TEST_LOG_OUTPUT pointed to /tmp/gotest.log which does not resolve through the MSYS2-to-native-Go boundary. Use relative path instead.


mannie-exe:
Hey there sorry I had no time to follow-up on this, I wanted to check-in and ask if these changes were needed and if I should make any changes:

https://github.com/cloudflare/cloudflared/commit/649705d2919824cadef4a5f1ce9624d2d65a5423#diff-a54ff182c7e8acf56acfd6e4b9c3ff41e2c41a31c9b211b2deb9df75d9a478f9R203

I can close this otherwise

jcsf:
Hi,
We don't allow external changes in the github workflows, for security reasons. We will be fixing it ourselves, when we have the time. Thank you for your contribution.

PR #1610, opened 2026-03-14
Fix: tunnel route dns creates CNAME in wrong zone when user has multiple domains

StringerBell69:
## Description
This PR fixes a bug in `cloudflared tunnel route dns` where the CLI would incorrectly create a CNAME record in the wrong zone if the account manages multiple domains.

### The Issue
Previously, the [RouteTunnel](cci:1://file:///Users/danielsumbo/Documents/workflow/cloudflared/cfapi/client.go:18:1-18:82) function blindly sent the route PUT request using the [ZoneID](cci:1://file:///Users/danielsumbo/Documents/workflow/cloudflared/credentials/credentials.go:29:0-31:1) embedded in the `cert.pem` upon login. If a user was logged in with `cert.pem` tied to `domain-A.com`, but ran `cloudflared tunnel route dns <tunnel> app.domain-B.com`, the API would attempt to create the DNS record inside the `domain-A.com` namespace (e.g. creating `app.domain-B.com.domain-A.com`).

### The Fix
1. Introduced [ListZones()](cci:1://file:///Users/danielsumbo/Documents/workflow/cloudflared/cfapi/client.go:19:1-19:29) in `cfapi` to correctly fetch all authorized zones attached to the user's account (leveraging [fetchExhaustively](cci:1://file:///Users/danielsumbo/Documents/workflow/cloudflared/cfapi/base_client.go:151:0-169:1) to avoid missing pagination records).
2. Modified [RouteTunnel](cci:1://file:///Users/danielsumbo/Documents/workflow/cloudflared/cfapi/client.go:18:1-18:82) to perform an in-memory suffix match (`strings.HasSuffix(route.Hostname(), "."+zone.Name) || route.Hostname() == zone.Name`) against the verified zones list from the API.
3. Once the matching zone is discovered, it dynamically constructs the correct endpoint (`/zones/<ZoneID>/tunnels/<id>/routes`) instead of falling back to the login certificate's default zone. 

Added comprehensive automated unit tests covering the exact matching, subdomain suffix mapping, and fallback logic if a zone fails to map.

### DIFF
--- a/cfapi/base_client.go
+++ b/cfapi/base_client.go
@@ -40,6 +40,7 @@ type baseEndpoints struct {
 	zoneLevel     url.URL
 	accountRoutes url.URL
 	accountVnets  url.URL
+	zones         url.URL
 }
 
 var _ Client = (*RESTClient)(nil)
@@ -60,7 +61,11 @@ func NewRESTClient(baseURL, accountTag, zoneTag, authToken, userAgent string, lo
 	}
 	zoneLevelEndpoint, err := url.Parse(fmt.Sprintf("%s/zones/%s/tunnels", baseURL, zoneTag))
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to create account level endpoint")
+		return nil, errors.Wrap(err, "failed to create zone level endpoint")
+	}
+	zonesEndpoint, err := url.Parse(fmt.Sprintf("%s/zones", baseURL))
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to create zones endpoint")
 	}
 	httpTransport := http.Transport{
 		TLSHandshakeTimeout:   defaultTimeout,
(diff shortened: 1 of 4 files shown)


claude-on-call:
Confirming this is still reproducible on **cloudflared 2026.8.2** (built 2026-08-14), six months after this PR was opened — the underlying issue in `RouteTunnel` still appears to be present.

We hit this twice on an account whose origin cert covers two zones (calling them `example.com` and `other-domain.net` here), running `cloudflared tunnel route dns <tunnel> service.example.com` each time:

- **Run 1:** created the correct CNAME in `example.com` (the tunnel worked) *and* a stray, non-functional record `service.example.com.other-domain.net` in the other zone. A double write, not just a wrong-zone write.
- **Run 2** (different tunnel, same account, same command shape): created *only* the stray record in `other-domain.net`. Nothing landed in `example.com` at all — the tunnel had no route and simply didn't resolve.

One more detail that might be useful for the fix and its tests: in both runs, the CLI's own log line claimed success and printed the wrong hostname regardless of what it actually did:

```
INF Added CNAME service.example.com.other-domain.net which will route to this tunnel tunnelID=<uuid>
```

That's true-looking output even on Run 2, where the real record was simply missing — so there's no visible signal from the command itself that DNS routing failed silently. Might be worth the fix also correcting that log line to reflect whichever zone the record actually landed in, or erroring if no zone matches.

This PR's approach — resolving via `ListZones()` and a suffix match, instead of trusting the login cert's zone — looks like it would fix all three symptoms above. It'd be great to see this reviewed; happy to test against it or provide redacted output if useful.

PR #1644, opened 2026-05-10
fix: isolate updater service test files

immanuwell:
## What

Updater service tests used `filepath.Join(os.TempDir(), "test")` as a shared fake binary path. If `/tmp/test` already exists as a dir, the package falls over before it tests updater logic. tiny papercut, but pretty easy to hit on a dev box.

This swaps it to `t.TempDir()` so each test gets its own scratch file. No prod code touched.

## Repro

```sh
mkdir -p /tmp/test
go test ./cmd/cloudflared/updater
```

Before: `open /tmp/test: is a directory`

After: updater tests pass.

## Tested

```sh
go test ./cmd/cloudflared/updater
go test -race -mod=vendor ./cmd/cloudflared/updater
go test ./...
```

`go test ./...` still hits local `ingress` ICMP permission failures here, unrelated to this change.

### DIFF
--- a/cmd/cloudflared/updater/workers_service_test.go
+++ b/cmd/cloudflared/updater/workers_service_test.go
@@ -23,8 +23,6 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-var testFilePath = filepath.Join(os.TempDir(), "test")
-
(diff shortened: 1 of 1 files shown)


Tsims61:
We to make the a The Bitcoin Terminal Exchange is extraordinary

PR #1672, opened 2026-06-09
Fix version typo in RELEASE_NOTES

jnoordsij:
(does not match commit message, but does match actual change)


jcsf:
Hi,
Thanks for your controbution but we do not change RELASE_NOTES file that is automatically managed by the release pipeline.

Issue #1365, opened 2024-11-30
🐛compile error


kinchan:
Hi, I used FreeBSD 14.
Compile error now.
Please help

----------------------------
gmake cloudflared

GOOS=freebsd GOARCH=amd64  go build -mod=vendor  -ldflags='-X "main.Version=2024.11.1-11-gb3304bf0" -X "main.BuildTime=2024-11-30-2155 UTC" ' github.com/cloudflare/cloudflared/cmd/cloudflared
# github.com/cloudflare/cloudflared/cmd/cloudflared/tunnel
cmd/cloudflared/tunnel/cmd.go:559:15: undefined: diagnostic.NewSystemCollectorImpl
gmake: *** [Makefile:138: cloudflared] Error 1

----------------------------------------

vrtareg:
I am on TrueNAS Core Jail with FreeBSD 13.4-RELEASE-p2 and having similar issues with latest and 2024.12.2 tag

```
gmake cloudflared
GOOS=freebsd GOARCH=amd64  go build -mod=vendor  -ldflags='-X "main.Version=2024.12.2-1-g2714d10d" -X "main.BuildTime=2025-01-05-2240 UTC" ' github.com/cloudflare/cloudflared/cmd/cloudflared
# github.com/cloudflare/cloudflared/diagnostic
diagnostic/diagnostic.go:170:30: undefined: network.NetworkCollectorImpl
gmake: *** [Makefile:138: cloudflared] Error 1
```

vrtareg:
Switching to 2024.11.1 tag helped to build.

vrtareg:
This looks like related to folder "diagnostic" which has been added after "2024.11.1" tag.
Is it even compatible with FreeBSD?

NothingTooSerious:
Hi,
The diagnostic code seems to support the official builds only and burns on all the coded but unofficially supported OS like the BSDs. By adding a FreeBSD flavour to the compile and duplicating the Mac SystemCollectorImpl implementation resulted in a successful compile. These steps were tested on the 2025.02.0 release. I would not use the current FreeBSD port version of cloudflared.

Make sure to utilise the Cloudflare GO fork installed using .teamcity/install-cloudflare-go.sh else unsupported crypto messages will prevent connection to CF.

* Patch 1

~~~
--- diagnostic/network/collector_unix.go.orig   2025-02-19 03:35:58 UTC
+++ diagnostic/network/collector_unix.go
[...]

PR #1707, opened 2026-08-02
Fix FreeBSD build by defining createTokenFile

kinchan:
Need issues #1365 patch first.

## Summary

Building `cloudflared` on FreeBSD currently fails with:

```text
cmd/cloudflared/common_service.go:56:12: undefined: createTokenFile
```

This worked around mid-July 2026, but no longer builds on the current HEAD.

## Environment

- OS: FreeBSD 14 and 15 (amd64)
- Build command:

```sh
gmake cloudflared
```

## Cause

`createTokenFile` is defined for Linux, macOS, and Windows, but there is no corresponding definition for FreeBSD.

Current implementations:

- `cmd/cloudflared/linux_service.go`
- `cmd/cloudflared/macos_service.go`
- `cmd/cloudflared/windows_service.go`

FreeBSD is missing an equivalent definition.

## Fix

Adding the following file restores the build successfully:

`cmd/cloudflared/freebsd_service.go`

```go
//go:build freebsd

package main

var createTokenFile = createTokenFileUnix
```

## Verification

After adding the file:

- `gmake cloudflared` completes successfully.
- The resulting `cloudflared` binary starts correctly.
- Basic tunnel operation has been verified on FreeBSD.

This change only affects FreeBSD builds and does not modify behavior on other platforms.

### DIFF
--- a/cmd/cloudflared/freebsd_service.go
+++ b/cmd/cloudflared/freebsd_service.go
@@ -1,0 +1,5 @@
+//go:build freebsd
+
+package main
+
+var createTokenFile = createTokenFileUnix


agoodkind:
Bump on this!

PR #1712, opened 2026-08-06
fix: bump quic-go fork pin to include the CVE-2025-59530 fix

### LINKED ISSUE #1709: replace directive on quic-go shadows the fix for CVE-2025-59530
go.mod pins the QUIC stack to a fork based on quic-go v0.45:

```text
require github.com/quic-go/quic-go v0.52.0
replace github.com/quic-go/quic-go => github.com/chungthuang/quic-go v0.45.1-0.20250428085412-43229ad201fd
```

The replace has no version on its left side, so it wins and the require line above it does nothing. That shadows every quic-go release since v0.45, including the fix for [CVE-2025-59530](https://github.com/advisories/GHSA-47m2-4cr7-mhcw) (7.5), which is patched in 0.49.1, 0.54.1 and 0.55.0. The pinned base is older than all three, so as far as I can tell no released cloudflared carries the fix.

It's a client-side crash. A server that sends HANDSHAKE_DONE before the handshake actually finishes makes the client drop Handshake keys before Initial keys, and the next undecryptable packet then trips the "no packets are queued after completion of the handshake" assert. cloudflared is the client here, so it takes something acting as the edge to hit it, but it's still a remote panic in the connection path.

I see the v0.59.1 bump landed twice and was reverted twice, both times shortly before a release:

- [`68620ef`](https://github.com/cloudflare/cloudflared/commit/68620efbce4c) landed 2026-06-12, reverted by [`2bcaf09`](https://github.com/cloudflare/cloudflared/commit/2bcaf0973404) on 2026-06-18 13:30. 2026.6.1 went out at 14:40.
- [`02eb75b`](https://github.com/cloudflare/cloudflared/commit/02eb75b56d29) landed 2026-06-18 18:20, reverted by [`a53d9e5`](https://github.com/cloudflare/cloudflared/commit/a53d9e5d44aa) on 2026-07-08 10:38. 2026.7.0 went out at 13:10.

Neither revert records a reason, so I won't guess at one. From outside the only visible thing is that v0.59 is a big move: the tracing and metrics code under `quic/` is written against `logging.PacketType` and `quic.Connection`, and both moved.

If the blocker is the size of that move rather than the fix itself, the fix backports onto the current pin by itself. `handleHandshakeConfirmed` in the v0.45 fork is the same as today's upstream apart from the receiver name and the time type, so [quic-go#5354](https://github.com/quic-go/quic-go/pull/5354) is seven lines:

```diff
diff --git a/connection.go b/connection.go
--- a/connection.go
+++ b/connection.go
@@ -851,6 +851,13 @@ func (s *connection) handleHandshakeComplete(now time.Time) error {
 }
 
 func (s *connection) handleHandshakeConfirmed(now time.Time) error {
+	// Drop initial keys.
+	// On the client side, this should have happened when sending the first Handshake packet,
+	// but this is not guaranteed if the server misbehaves.
+	// See CVE-2025-59530 for more details.
+	if err := s.dropEncryptionLevel(protocol.EncryptionInitial, now); err != nil {
+		return err
+	}
 	if err := s.dropEncryptionLevel(protocol.EncryptionHandshake, now); err != nil {
 		return err
 	}
```

Dropping Initial keys twice is safe on this base, which matters because the normal client path already dropped them by that point. `sentPacketHandler.DropPackets` returns early when the packet number space is already gone, `DiscardInitialKeys` is nil-safe, and `cryptoStream.Finish` only errors when the crypto stream still has unread data. Upstream relies on the same property, its own test asserts `DiscardInitialKeys` is called twice, and upstream's `dropEncryptionLevel` has no guard the v0.45 fork lacks.

I applied it to the pinned revision and built and tested a project that embeds the tunnel transport in-process. Build and the tunnel transport tests stay green. I could not run quic-go's own suite against it, a vendored copy has no test files, so that needs a run inside the fork.

I can open the PR against chungthuang/quic-go so the pin can move to a new revision of the same base, or you can just take the diff, it needs no attribution. This is meant as an interim step and the v0.59 upgrade still works whenever it's convenient, at which point this can be dropped.

stareezy-1:
## Summary

Fixes https://github.com/cloudflare/cloudflared/issues/1709

## Problem

The `replace` directive pins the QUIC stack to a fork based on quic-go v0.45, which predates the fix for **CVE-2025-59530** (GHSA-47m2-4cr7-mhcw, 7.5). A server that sends HANDSHAKE_DONE before the handshake actually finishes makes the client drop Handshake keys before Initial keys; the next undecryptable packet then trips the "no packets are queued after completion of the handshake" assert — a remote crash for cloudflared as the QUIC client.

Fuller `quic-go` upgrades (v0.59.1) were landed and reverted twice before releases, so this PR takes the backport path suggested in the issue analysis: the fork pin is bumped to a commit that applies quic-go#5354 onto the **exact previously-pinned fork state**. The only source change is the 7-line fix in `connection.go` (drop initial keys when the handshake is confirmed) — no API migration.

## Changes

- `go.mod` / `go.sum`: fork pin bumped to `stareezy-1/quic-go v0.0.0-20260806091559-d3ee2305a943` (pinned fork state + backport).
- `vendor/github.com/quic-go/quic-go/connection.go`: the backported fix (verified as the only vendored source change).
- Companion upstream PR: https://github.com/chungthuang/quic-go/pull/16 — once an equivalent commit lands on the fork's `cloudflare` branch, the replace should be flipped back to `chungthuang/quic-go` at that commit.

## Verification

- `go build ./...` passes.
- `go test ./quic/...` passes (both packages).
- The vendored diff vs the previous pin is exactly the 7-line fix.

## Notes

- The pin currently points at a personal fork commit solely because the backported commit does not exist on `chungthuang/quic-go` yet; it carries the identical module path and fork state.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -109,4 +109,4 @@ replace github.com/prometheus/golang_client => github.com/prometheus/golang_clie
 replace gopkg.in/yaml.v3 => gopkg.in/yaml.v3 v3.0.1
 
 // This fork is based on quic-go v0.45
-replace github.com/quic-go/quic-go => github.com/chungthuang/quic-go v0.45.1-0.20250428085412-43229ad201fd
+replace github.com/quic-go/quic-go => github.com/stareezy-1/quic-go v0.0.0-20260806091559-d3ee2305a943
--- a/go.sum
+++ b/go.sum
@@ -9,8 +9,6 @@ github.com/bytedance/sonic/loader v0.2.0 h1:zNprn+lsIP06C/IqCHs3gPQIvnvpKbbxyXQP
 github.com/bytedance/sonic/loader v0.2.0/go.mod h1:ncP89zfokxS5LZrJxl5z0UJcsk4M4yY2JpfqGeCtNLU=
 github.com/cespare/xxhash/v2 v2.3.0 h1:UL815xU9SqsFlibzuggzjXhog7bL6oX9BbNZnL2UFvs=
 github.com/cespare/xxhash/v2 v2.3.0/go.mod h1:VGX0DQ3Q6kWi7AoAeZDth3/j3BFtOZR5XLFGgcrjCOs=
-github.com/chungthuang/quic-go v0.45.1-0.20250428085412-43229ad201fd h1:VdYI5zFQ2h1/qzoC6rhyPx479bkF8i177Qpg4Q2n1vk=
-github.com/chungthuang/quic-go v0.45.1-0.20250428085412-43229ad201fd/go.mod h1:MFlGGpcpJqRAfmYi6NC2cptDPSxRWTOGNuP4wqrWmzQ=
 github.com/cloudflare/backoff v0.0.0-20240920015135-e46b80a3a7d0 h1:pRcxfaAlK0vR6nOeQs7eAEvjJzdGXl8+KaBlcvpQTyQ=
 github.com/cloudflare/backoff v0.0.0-20240920015135-e46b80a3a7d0/go.mod h1:rzgs2ZOiguV6/NpiDgADjRLPNyZlApIWxKpkT+X8SdY=
 github.com/cloudwego/base64x v0.1.4 h1:jwCgWpFanWmN8xoIUHa2rtzmkd5J2plF/dnLS6Xd/0Y=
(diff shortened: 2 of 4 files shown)


lexfrei:
The fork PR looks right, but this one repoints the replace at a personal fork. That moves the trust anchor for cloudflared QUIC stack off the chungthuang fork, and the build breaks if that repo is ever renamed or removed. Once chungthuang/quic-go#16 is merged, this can just bump the existing pin to the new commit on that fork, which is a one-line change and needs no new dependency source.

stareezy-1:
Agreed — the replace should stay anchored on the chungthuang fork. Closing this; once chungthuang/quic-go#16 merges, the existing pin can be bumped to the new commit on that fork (one-line change, no new dependency source).

stareezy-1:
Closing per review: repointing the quic-go replace at a personal fork is not acceptable; the fix will land by bumping the existing chungthuang pin after chungthuang/quic-go#16 merges.

PR #1735, opened 2026-09-07
Fix Windows self-update reporting success while leaving cloudflared.exe missing

### LINKED ISSUE #667: Updates from Windows with `cloudflared.exe update` can fail 🐛
**Describe the bug**
On Windows cloudflared.exe, (sometimes) does not updates with command `cloudflare.exe update`. 

**To Reproduce**
1. Download cloudflared.exe from github to C:\Cloudflared\bin\cloudflared.exe
3. Within a terminal cmd, run `cd C:\Cloudflared\bin` and then `cloudflared.exe update`
4. Sometimes it works, sometimes it fails

When it fails, cloudfared.exe does not exists anymore, but there is 3 "new" files : 
- cloudflared.exe.old
- cloudflared.exe.new
- cfd_update.bat

**Expected behavior**
Cloudflared is updated and service is up

**Environment and versions**
 - OS: Windows 2019 Standard edition (1809)
 - Architecture: Intel x86_64
 - Version:  2022.5.3 (built 2022-05-30-1513 UTC) from github

**Logs and errors**
No errors, it says : 
```
2022-06-16T13:48:52Z INF cloudflared has been updated version=2022.6.1
cloudflared has been updated to version 2022.6.1
```
And return code is 11

but cloudflare.exe does not exists anymore and service is down

NAVEENKUMARKR777:
## Summary

Fixes #667, a 4-year-old, maintainer-confirmed bug where `cloudflared.exe update` on Windows can report success (`cloudflared has been updated`, exit code 11) while actually leaving `cloudflared.exe` deleted and the service down, with `cloudflared.exe.old`/`cloudflared.exe.new`/`cfd_update.bat` littered on disk.

Root cause, confirmed by re-reading the generated `cfd_update.bat` against the current code (community member @JonathonFS independently traced the same chain with Process Monitor in the issue thread):

1. **`sc stop` doesn't wait.** It only requests a stop and returns immediately. If the service (or the very `update` process, which is running the same binary) hadn't released the file yet, the following `rename` raced it.
2. **`runWindowsBatch` blocked on the whole script via `cmd.Output()`**, including the trailing `sc start`. That kept the `cloudflared.exe update` process alive for the entire script — but the script's job is to rename *that exact running binary*, and Windows won't complete that rename while something is still waiting on the process. The rename fails right when the tool is depending on it most.
3. **The script always ended with an unconditional `exit /b 0`.** Any rename failure inside the script was swallowed, so the Go side always saw "success" regardless of what actually happened on disk.

Together these produce exactly the reported symptom: `.old`/`.new` left behind, `cloudflared.exe` missing, service down, log says it worked.

## Fix

- Use `net stop` instead of `sc stop` — it blocks until the service has actually stopped.
- Switch `runWindowsBatch` to `cmd.Start()` instead of `cmd.Output()`, so the update process returns (and the CLI framework's `ExitCoder` handling exits it) right after launching the script, instead of holding the binary open for the script's whole run.
- Check the errorlevel after each `rename`; if replacing the binary fails partway through, restore the original from `.old` rather than leaving no executable at all.
- Have the script delete itself (`del "%~f0"`) once done, since the Go side can no longer safely `os.Remove` the batch file after switching away from the blocking call.
- Render the script with CRLF line endings — `cmd.exe`'s parser is unreliable with GOTOs/labels over Unix line endings, which the new error-handling branches now rely on.

## Test plan

- [x] `go test ./cmd/cloudflared/updater/...` — full package suite passes, including two new tests covering the rendered script's line endings and control flow (`net stop` present/`sc stop` absent, checked renames, rollback path, self-cleanup).
- [x] `go build ./...` on the host platform.
- [x] `GOOS=windows GOARCH=amd64 go build ./...` — full `cloudflared` binary cross-compiles cleanly.
- [ ] Not able to exercise the actual Windows service stop/rename/restart sequence end-to-end in this environment (no Windows host available) — would appreciate a maintainer or an affected reporter from #667 validating on a real Windows install.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/cmd/cloudflared/updater/workers_update.go
+++ b/cmd/cloudflared/updater/workers_update.go
@@ -2,6 +2,7 @@ package updater
 
 import (
 	"archive/tar"
+	"bytes"
 	"compress/gzip"
 	"errors"
 	"fmt"
@@ -13,6 +14,7 @@ import (
 	"path"
 	"path/filepath"
 	"runtime"
+	"strings"
 	"text/template"
 	"time"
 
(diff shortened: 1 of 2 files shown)


oga35767-eng:
Yu

PR #1741, opened 2026-09-14
Fix websocket.Conn.Read dropping the tail of messages larger than the read buffer

pkit:
## Summary

For `tcp://`, `ssh://`, `rdp://`, `smb://`, bastion and socks-over-WebSocket ingress services, cloudflared unwraps the client's WebSocket frames itself via the server-side `websocket.Conn`. Its `Read` (`websocket/connection.go`) called `wsutil.ReadClientBinary`, which returns one complete WebSocket message, and then returned `copy(reader, data)`. Any bytes of the message past `len(reader)` were silently discarded.

`stream.Pipe` drives that `Read` through `cfio.Copy`, which uses a pooled 16 KiB buffer (`cfio/copy.go`). So every client message larger than 16 KiB reached the origin truncated to its first 16 KiB, while the origin-to-client direction was unaffected. The origin then saw a corrupted byte stream (in our case a broken TLS record) and reset the connection, which shows up in debug logs as `downstream->upstream copy: stream error: stream ID N; NO_ERROR`.

The code has been this way since e2262085 (TUN-3617, 2020). `cloudflared access tcp` never triggers it because it also copies through `cfio.Copy` and therefore never emits messages larger than 16 KiB, but any other WebSocket client that writes larger binary messages does.

## Reproduction

- Ingress rule `tcp://<host>:443` behind Cloudflare Access.
- A client that dials the `wss://` endpoint and forwards a TCP stream as one binary message per `Write` (for example `io.Copy` from a `*net.TCPConn`, which produces 32 KiB messages).
- Client to origin: of each message exactly the first 16 KiB arrives; a 4 MiB stream sent as 32 KiB messages arrives as every other 16 KiB chunk (first difference at byte 16385). With 64 KiB messages still only 16 KiB per message survives. With messages of 16 KiB or less everything arrives intact.
- Same result with 2025.5.0 on http2 and 2026.8.2 on QUIC, so it does not depend on the tunnel transport.

## Fix

Keep the unread remainder of a message in a `bytes.Buffer` on `Conn` and drain it on subsequent `Read` calls before reading the next message. This mirrors what `GorillaConn.Read` in the same file already does. `Write`, ping/pong handling, control-frame handling and error behaviour are unchanged.

## Tests

`websocket/connection_test.go` adds:

- `TestConnReadMessageLargerThanBuffer`: a single 100 KiB client binary message read through 16 KiB `Read`s arrives byte for byte.
- `TestConnReadMixedMessageSizes`: messages of 1, 16383, 16384, 16385, 40000, 49152, 102400, 7 and 16384 bytes sent back to back arrive concatenated in order.

Both tests fail on the previous code (the reader times out waiting for the dropped tail) and pass with the fix. `go test ./websocket/` and `go vet ./websocket/` are clean. The change applies unmodified on 2025.5.0, 2026.8.2 and master.

C✳


websocket/connection.go
```diff
@@ -81,6 +81,9 @@ func (c *GorillaConn) SetDeadline(t time.Time) error {
 type Conn struct {
 	rw  io.ReadWriter
 	log *zerolog.Logger
+	// readBuf holds the unread remainder of the last message when it was larger than the buffer passed to Read.
+	// Without it, everything past len(reader) of a message would be silently dropped.
+	readBuf bytes.Buffer
```
jcsf:
```suggestion
	readBuf []byte
```

websocket/connection.go
```diff
@@ -99,11 +102,22 @@ func NewConn(ctx context.Context, rw io.ReadWriter, log *zerolog.Logger) *Conn {
 
 // Read will read messages from the websocket connection
 func (c *Conn) Read(reader []byte) (int, error) {
+	// Intermediate buffer may contain unread bytes from the last read, start there before blocking on a new frame
+	if c.readBuf.Len() > 0 {
+		return c.readBuf.Read(reader)
```
jcsf:
```suggestion
if len(c.readBuf) > 0 {
	n := copy(reader, c.readBuf)
	c.readBuf = c.readBuf[n:]
	if len(c.readBuf) == 0 {
		c.readBuf = nil
	}
	return n, nil
```

websocket/connection.go
```diff
@@ -99,11 +102,22 @@ func NewConn(ctx context.Context, rw io.ReadWriter, log *zerolog.Logger) *Conn {
...
 		return 0, err
 	}
-	return copy(reader, data), nil
+
+	copied := copy(reader, data)
+
+	// Write unread bytes to readBuf; if everything was read this is a no-op
+	c.readBuf.Write(data[copied:])
+
+	return copied, nil
```
jcsf:
```suggestion
    n := copy(reader, data)
    if n < len(data) {
  		c.readBuf = data[n:]
    }
   return n, nil
```

pkit:
@jcsf it looks like the reader accepts messages of any length (with or without this patch)
Which probably needs a follow up to fix.

